PR #16, opened 2020-08-08
cli module for git users

andreeatoma:
custom React Component which handles Git commands


frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,63 @@
...
+        } else {
+            this.setState({
+                message: 'try again'
+            });
+        }
+    }
+    
+    render() {
+        return (
+            <form onSubmit={this.handleSubmit}>
```
iampava, line 45: Great job for using a `<form>` + `<button type="submit">`. 

On the other hand, why 2 event listeners: `onSubmit` and `onClick`? We just need the `onSubmit`

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,63 @@
...
+                <input 
+                    type="text" 
+                    value={this.state.value}
+                    onChange={this.handleInputChange}
```
iampava, line 49: Let's not update the state every time the user types, it's not necessary. We only care about the value when the user submits the form.

So, let's take the final result from the event handler of the `onSubmit` function. To do this you'll need to add a `name` property to the `<input />`

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,63 @@
...
+        event.preventDefault();
+    }
+    handleInputChange(event) {
+        this.setState({
+            value: event.target.value
+        });
+    }
+
+    handleMatch() {
+        let inputValue = this.state.value;
```
iampava, line 26: Use destructuring pattern. 

`const {a, b, c} = this.props` 

instead of

```
const a  = this.props.a;
const b = this.props.b;
const c = this.props.c;
```

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,63 @@
...
+        let defaultValue = this.props.defaultText;
```
iampava, line 27: Please change to just one prop which can be either an Array or a string. You should check the type of this props and based on it do the rest of the logic.

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,63 @@
...
+class MatchWords extends React.Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            value: '',
+            message: '',
+        }
+
+        this.handleInputChange = this.handleInputChange.bind(this);
```
iampava, line 12: Bind is ok but with a lot of methods things get a little messy.

I suggest you declare the custom methods on the class like this:

```
handleSubmit  = (event) => {
  // ...
}
```

This way **this** will not get lost.

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,100 @@
...
+//     }
+
+
+//     render() {
+//         const { message } = this.state;
+//         return (
+//             <form onSubmit={this.handleSubmit}>
+//                 <input
+//                     type="text"
+//                     name="input_text"
```
iampava, line 46: The name is too generic. Usually we want it to reflect what's in the input. For example `password` for password inputs, `email` for emails, etc. Here, let's call it something related to Git, since you're typing GIT commands. Maybe `git_command`, `cli_command` etc.

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,100 @@
...
+
+//         this.state = {
+//             message: '',
+//         }
+//     }
+
+//     handleSubmit = (event) => {
+//         event.preventDefault();
+
+//         const inputValue = document.querySelector('[name="input_text"]').value;
```
iampava, line 15: Let's not use `querySelector` on the `document`. It's highly ineffective searching the entire DOM. You can get directly to the input from the `event`, no querySelectors.

`event.target['input_text'].value` should work

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,100 @@
...
+//         const {defaultValue} = this.props;
+
+//         if (Array.isArray(defaultValue)) {
+//             var result = defaultValue.some(
```
iampava, line 19: Why 2 variables, `result` and `wordToMatch`? 

You should have just one, and depending on the situation (`Array` or `string`) you give it a certain value. Also, let's use `let` instead of `var`.

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,100 @@
+import React, { useState } from 'react';
+
+// class MatchWords extends React.Component {
```
iampava, line 3: Let's have 2 different files, instead of commenting one component.

frontend-ssr/components/cli_git/match_commands.js
```diff
@@ -0,0 +1,100 @@
...
+//                 </button>
+//                 <br />
+//                 {message}
+//             </form>
+//         )
+//     }
+// }
+
+function MatchWords(props) {
+    let [message, setMessage] = useState('');
```
iampava, line 60: Great use of hooks! Same comments as above should apply here as well.

PR #18, opened 2020-08-11
Adding login backend 

SirCQQ:
Backend:

- Create Login route => returns a JWT (needs .env with TOKEN_SECRET variable) 

- Create Register route

- Create middleware function to check if the jwt token is still valid (7 days availability) for future requests that required authentication

- Create Model for Users in DB (MongoDB Atlas) (need .env with DB_CONNECTION_LINK) 

- Adding schema for validation of the register and login forms 

# Ignore the .TSX created future work in progress


frontend-ssr/models/User.schema.js
```diff
@@ -0,0 +1,35 @@
+const mongoose = require("mongoose");
```
iampava, line 1: Don't forget to configure eslint and run it for all files.

frontend-ssr/package-lock.json
```diff
@@ -0,0 +1,8758 @@
+{
```
iampava, line 1: NextJS uses `yarn` so we should install dependencies this way. YARN is another package manager, similar to `npm`. Run `npm install -g yarn` to install it globally and then `yarn` inside the project to install de dependencies.

Also delete this file, instead we should have a `yarn-lock.json` or smth like that.

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,38 @@
...
+const User = require("../../../models/User.schema");
+const bcrypt = require("bcryptjs");
+const jwt = require("jsonwebtoken");
+const { loginValidation } = require("../../../utils/validation/userValidation")
+
+//Vars
+const error_message = { status: "fail", message: 'Email or Password is incorrect' }
+
+export default async (req, res) => {
+    let { method, body } = { ...req };
```
iampava, line 11: Why are you destructuing `req` as well? 

This `let { method, body } = req` should be enough.

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,38 @@
...
+    switch (method) {
+        case "POST":
+            //destruct the req
+            let { email, password } = { ...body };
```
iampava, line 15: Same as above with the destructuring. Please change in all places where you're using like this.

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,38 @@
+//Imports
+const User = require("../../../models/User.schema");
+const bcrypt = require("bcryptjs");
+const jwt = require("jsonwebtoken");
+const { loginValidation } = require("../../../utils/validation/userValidation")
+
+//Vars
+const error_message = { status: "fail", message: 'Email or Password is incorrect' }
```
iampava, line 8: Let's name this something a little less generic. `error_message` doesn't tell you what kind of error or situation are we in.

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,38 @@
...
+            let { email, password } = { ...body };
+            let { error } = loginValidation({ email, password })
+            //Check valid form 
+            if (error) return res.status(400).json(error_message);
+            //Check user base on username or email
+            let user = await User.findOne({ $or: [{ username: email }, { email: email }] });
+            if (!user) return res.status(400).json(error_message);
+            //User exists => Check password  
+            const validPass = await bcrypt.compare(password, user.password);
+            if (!validPass) return res.status(400).json(error_message);
```
iampava, line 24: Let's give more specific error messages. If username or email doesn't exist when logging it, let's say that. Otherwise, if the password is wrong let's give a specific message regarding this.

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,38 @@
...
+    let { method, body } = { ...req };
+    switch (method) {
+        case "POST":
+            //destruct the req
+            let { email, password } = { ...body };
+            let { error } = loginValidation({ email, password })
+            //Check valid form 
+            if (error) return res.status(400).json(error_message);
+            //Check user base on username or email
+            let user = await User.findOne({ $or: [{ username: email }, { email: email }] });
```
iampava, line 20: What if this one fails? We should still return a 500 message with something like "Oups, something went wrong. Try again.". 

Question: do we have to do this for all calls to the DB or can we implement a generic catch for fails like this?

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,38 @@
...
+            if (!user) return res.status(400).json(error_message);
+            //User exists => Check password  
+            const validPass = await bcrypt.compare(password, user.password);
+            if (!validPass) return res.status(400).json(error_message);
+            //After checking password update the last login
+            let new_entry = await User.findOneAndUpdate({ _id: user._id }, { $set: { lastLogin: Date.now() } }, { new: true })
```
iampava, line 26: Same here, what if the update fails?

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,38 @@
...
+            console.log(new_entry);
+            // Create jwt
+            const token = jwt.sign({ _id: user._id }, process.env.TOKEN_SECRET, { expiresIn: '7d' });
+            res.json({ status: "success", auth_token: token });
```
iampava, line 30: Shouldn't we return the user data after login as well?

frontend-ssr/pages/api/user/register.js
```diff
@@ -0,0 +1,56 @@
+//Imports 
+
+const User = require("../../../models/User.schema");
+import dbConnect from "../../../utils/dbConnect";
+const bcrypt = require("bcryptjs")
+dbConnect();
```
iampava, line 6: I see you're not running `dbConnect` from `login.js`. Is this intended or you forgot about it?

frontend-ssr/pages/api/user/register.js
```diff
@@ -0,0 +1,56 @@
...
+const {registerValidation} = require("../../../utils/validation/userValidation") 
+// import UserRequests from "../../../utils/user.requests";
+
+export default async (req, res) => {
+    let { method, body } = { ...req };
```
iampava, line 11: Destructuring. Same feedback as in `/login` route.

frontend-ssr/pages/api/user/register.js
```diff
@@ -0,0 +1,56 @@
...
+            // console.log(method, body);
+            let {name,username,avatar,email,password}={...body}
+            let {error} = registerValidation(body);
+            if (error)return res.status(400).json({"message":error})
+            let errors=[]
+            // Encrypt the password 
+            const salt= await bcrypt.genSalt(10); 
+            const hash_password= await bcrypt.hash(password,salt)
+            //Create the user for the database 
+            let created_user=new User({
```
iampava, line 25: If the key and variable name are the same, we can use the shorthand version

```

new User({
  name,
  username,
  avatar, 
  email,
  password: hash_password
})
```

frontend-ssr/pages/api/user/register.js
```diff
@@ -0,0 +1,56 @@
...
+                name:name,
+                username:username,
+                avatar:avatar,
+                email:email,
+                password:hash_password
+            });
+            //Check if the email and username already exists
+            const emailExists=await User.findOne({email:body.email});
+            if(emailExists){ errors.push({email:"Email already exists"});}
```
iampava, line 34: In the `login` route you extracted the error messages in their separate variables. I think that approach is better so let's do this here as well. 

Also, same feedback regarding `await try-catch'es`. Let's handle them

frontend-ssr/pages/api/user/register.js
```diff
@@ -0,0 +1,56 @@
...
+            if(usernameExists) {errors.push({username:"Username already exists"});}
+            if(errors.length>0)return res.status(400).json({status:'fail',errors:errors})
+            //Try to save the user into the database 
+            try{
+                let user=await created_user.save();
+                res.json(user);
+            }
+            catch(err){
+                //catch the fail to save to the database 
+                res.status(400).json({'message':err});
```
iampava, line 45: If it fails here it should not be a 400 but a 500, right?

frontend-ssr/utils/validation/userValidation.js
```diff
@@ -0,0 +1,28 @@
+const Joi= require("@hapi/joi");
```
iampava, line 1: Never heard/used `@hapi/joi` before, but it seems the package has moved: https://github.com/sideway/joi/issues/2411

Also, is this they best way to validate? Doesn't mongoose offer something similar?

frontend-ssr/pages/api/user/check.js
```diff
@@ -0,0 +1,7 @@
+const {validate_jwt} =require("../../../utils/validation/middleware.validation")
```
iampava, line 1: Not sure what this route does...🤔

frontend-ssr/utils/validation/middleware.validation.js
```diff
@@ -0,0 +1,18 @@
...
+const error_message={status:"fail",message:"Invalid Token"}
+
+module.exports.validate_jwt=async(req)=>{
+    //Grab the auth-token from headers
+    const token = req.headers["auth-token"];
+    //No header return false 
+    if(!token) return false;
+    try{
+        //Verify header
+        const verify=await jwt.verify(token,process.env.TOKEN_SECRET);
```
iampava, line 11: If it's valid, let's return the data encrypted in the JWT.

frontend-ssr/pages/api/user/login.js
SirCQQ: That's a good question, I will look into it for a generic error handling for DB connection.

frontend-ssr/pages/api/user/login.js
SirCQQ: I think this is more like a preference, I use to return just the `user._id` and check it if necessary, but we can return the full user in the JWT

frontend-ssr/pages/api/user/register.js
SirCQQ: I forgot about it, resolve

frontend-ssr/utils/validation/userValidation.js
SirCQQ: I will look into it. Eventually returning with a mongoose version or own light `form` validation.

frontend-ssr/pages/api/user/check.js
SirCQQ: Change to `~/utils/...` . If it's not that DM me to talk about and I will update it

frontend-ssr/pages/api/user/register.js
SirCQQ: Surrounded with try-catch and moved error to separate variables to handle errors

frontend-ssr/.gitignore
```diff
@@ -22,9 +22,11 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+yarn.lock*
```
iampava, line 25: We must not ignore this one. The `yarn.lock` file makes sure we all install the exact same dependencies, meaning exact same version. Otherwise , different `yarn` commands may have different end results in the `node_modules`.

frontend-ssr/package-lock.json
iampava: The `package-lock.json` is still here. Did you forget to `push` the new changes?

frontend-ssr/pages/api/user/login.js
iampava: Another mention I just remembed. Let's CAPS_LOCK constant names so we're all on the same page with the styling.

frontend-ssr/pages/api/user/register.js
iampava: Not sure what this confused emoij means 😅

frontend-ssr/pages/api/user/register.js
iampava: Thanks. Just a reminder to insert the CAPS_LOCK feedback here is well.

frontend-ssr/pages/api/user/check.js
iampava: Oooooh, now I saw this is an `api` route. Is it the `ping` one from the code I sent you? Or when is this route gonna be used?

frontend-ssr/utils/validation/userValidation.js
SirCQQ: Sooo, mongoose does have a form of validation with the schema as a did in the model with 
`{
name:{
type:String,
required:true
}}`

And you can do something like this with a small twist on the error handling 
`{
name:{
type:String,
required:[true,"Name is required"] 
}}`
This will create a custom error message inside the mongoose validation, if there is a custom message for every required field, we can collect all the errors inside the catch like 
` let save_errors={};
        for(let i in err.errors){
          save_errors[i]=err.errors[i].message;
        }`

The result looking like this 
`{
    "name": "Name is required",
    "username": "Username is required",
    "email": "Email is required"
}`

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,74 @@
...
+
+const BAD_PASSWORD_ERROR = {
+  status: "fail",
+  message: "Incorect password ",
+};
+
+const UPDATE_LAST_LOGIN_ERROR = {
+  status: "fail",
+  message: "Something went wrong. Please try again",
+};
```
andyrkn, line 30: im thinking the code would look nicer if you take the errors in a separate file, same for the other cases

frontend-ssr/pages/api/user/login.js
```diff
@@ -0,0 +1,74 @@
...
+
+//Route function
+export default async (req, res) => {
+  let { method, body } = req;
+  switch (method) {
```
andyrkn, line 35: instead of forcing a switch you can say if method != post =>  bad request and then you reduce nesting twice

iampava:
Cherry-picked the commits here and cleaned them up a little bit. Then created a new PR - https://github.com/iampava/frontend.ro/pull/45 that was merged. Closing this one

PR #21, opened 2020-08-26
validating git commands after code review

andreeatoma:


frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+      isMatch = defaultValue[currentCommandIndex] === gitCommand.trim().toLowerCase();
+    }
+
+    if (!isMatch) {
+      this.setState({
+        message: 'please try again',
+        buttonText: 'Next',
+      });
+    } else if (currentCommandIndex === defaultValue.length - 1) {
```
iampava, line 45: Shouldn't this be `defaultValue.length - 2`. We want to set the text to `Done` if you're on the last step. Right?

frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -0,0 +1,78 @@
...
+  currentCommandIndex: number,
+  message: string,
+  buttonText: string
+}
+
+export default class MatchCommandClassComponent extends React.Component<Props, State> {
+  constructor(props) {
+    super(props);
+
+    this.handleSubmit = this.handleSubmit.bind(this);
```
iampava, line 16: We can skip this step with the help of `babel` which is built-into this project. Just change the function declaration on line 25 to this:

```
handleSubmit  = (event) => { 
  // ...
}
```

frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+
+interface Props {
+  defaultValue:Array<RegExp> | Array<string>
```
iampava, line 4: Let's change the name of this prop. `defaultValue` is not meaningful enough. Let's call it "commands".

iampava: Also, `eslint` I think it should have complained about the lack of space between `defaultValue:` and `Array<RegExp`.

iampava: Also also, fyi: you can also do `RegExp[]` instead of `Array<RegExp`. But you don't have to change it, use whichever you like the most.

iampava (changes requested):

frontend-ssr/components/cli_git/git_commands_cc.tsx
andreeatoma: Ok. I'll do RegExp[].

frontend-ssr/components/cli_git/git_commands_cc.tsx
andreeatoma: It means that I set the text to Done, after the user types git commit -m "Some message". Not after git push?

iampava: Yes, this is how it should work. After git push you no longer show the button, it's done!

PR #23, opened 2020-09-03
new functionality on git submit component, and scss added

andreeatoma:


frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
+@import '../../styles/variables';
+
+.commands {
+    width: 35vw;
```
iampava, line 4: The best practice here is not to give the component a width, but to get it's width from it's parent, because we don't know where it will be used. So, let's give it 100% width, and restrict the parent size to `35vw`.

frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
...
+
+    .commands__cli {
+        height: 50vh;
```
iampava, line 7: Same as above. We cannot assume it's height because we don't know where it will be used. Same aproach, have full width/height of the parent.

frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
...
+        overflow: auto;
+        padding: 2vh 1vw;
```
iampava, line 9: On mobile devices this padding might not be big enough. Let's use `em` from now on regarding sizes. If you've never used it, read about it from MDN.

frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
...
+    opacity: 0; 
+  } 
+  50%{ 
+    opacity: 1; 
+  } 
+  100%{ 
+    opacity: 0; 
+  } 
+} 
+@media (max-width: 720px) {
```
iampava, line 68: No need for this media query with the new approach to width/height.

frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
...
+                display: inline-block;
+        }
+    }
+    
+}
+.btn {
+    visibility: hidden;
+}
+ 
+@keyframes animate{
```
iampava, line 57: We may need this animation in other parts as well. So let's externalize it. Maybe create an `animations.scss` file and import it from `index.scss`. Also, let's change it's name. `animate` is WAY WAY to generic.

frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -1,22 +1,66 @@
 import React from 'react';
```
iampava, line 1: Let's rename the file. If the styles are called `GitCli.module.scss` then the component file should be called `GitCli.tsx`.

frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
...
+        p {
+            margin: 0;
+        }
+    }
+}
+.form {
+    position: relative;
+    color: color(white);
+    .form__input {
+        width: 100%;
```
iampava, line 30: Please add `color: color(white)`. The text is hard to read when you're writing.

frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
...
+        background-color: color(gray);
+        padding: 1em;
+        border: none;
+        &:focus ~ span::after {
+            width: 0;
+        }
+    }
+    span {
```
iampava, line 38: I was confused who this `span` is at first. Let's add a class to it, so we know it's the blinking cursor.

frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -26,50 +70,79 @@ export default class MatchCommandClassComponent extends React.Component<Props, S
...
+            onSubmit={this.handleSubmit}
+          >
+            <input
+              className={styles.form__input}
+              type="text"
+              name="gitCommand"
+              required
+            />
+            <span />
+            <button className={styles.btn} type="submit">
```
iampava, line 140: This button is hidden. It should be visible, in case someone sends from a tablet/mobile. Also, rembmer the button should change it's text, as it does now.

frontend-ssr/components/cli_git/GitCli.module.scss
```diff
@@ -0,0 +1,72 @@
...
+        }
+        p {
+            margin: 0;
+        }
+    }
+}
+.form {
+    position: relative;
+    color: color(white);
+    .form__input {
```
iampava, line 29: No need to create a class here. You can just style directly after `input[type="text"]` as we know its just one.

iampava (changes requested):

iampava:
Overall the biggest concern is how you insert/create the command messages. You should create them on the fly, meaning in the `handleSubmit` function. Not have them previously in an array. You get the text from the form, the type depending on the `isMatch` property, and calculate the `timestamp` there as well.

frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -1,22 +1,66 @@
 import React from 'react';
+import { v4 as uuidv4 } from 'uuid';
```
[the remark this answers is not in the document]
andreeatoma: I can't find the folder /services. Do I have to create a folder called services with a file called **Utils.ts** or the folder already exists?

frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -1,22 +1,66 @@
...
+enum MessageType {
+  COMMAND = 'command',
+  ERROR = 'error',
+  SUCCESS = 'success'
 }
 
-export default class MatchCommandClassComponent extends React.Component<Props, State> {
-  constructor(props) {
+const date = new Date();
+const timestamp = date.toLocaleTimeString();
```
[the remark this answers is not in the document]
andreeatoma: Yep. I didn't notice, but now that I see it, it makes a lot of sense.

frontend-ssr/components/cli_git/git_commands_cc.tsx
```diff
@@ -1,22 +1,66 @@
...
+      commandMessages: [{
+        text: '', type: MessageType.COMMAND, time: timestamp, id: uuidv4(),
+      },
+      {
+        text: '', type: MessageType.COMMAND, time: timestamp, id: uuidv4(),
+      },
+      {
+        text: '', type: MessageType.COMMAND, time: timestamp, id: uuidv4(),
+      }],
+      successMessages: [{
```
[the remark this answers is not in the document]
andreeatoma: Ok.

frontend-ssr/components/cli_git/GitCli.module.scss
andreeatoma: I've used `em` in my previous projects. Thanks!

frontend-ssr/components/cli_git/git_commands_cc.tsx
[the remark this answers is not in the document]
iampava: No, it doesn't. You have to create it yourself

frontend-ssr/styles/_variables.scss
```diff
@@ -2,7 +2,13 @@ $colors: (
     white: #fff,
     darker-white: #f8f8f8,
     black: #1b1b15,
-    red: #eb4d4b,
+    green: #58af58,
+    grey: #77838b,
+    silver: #bcc2c6,
+    yellow: #f7df1e,
+    danger: #eb4d4b,
```
iampava, line 9: We decided to call it `red` in this app, so you should have removed `danger` and kept the existing one, to not break other parts of the app.

frontend-ssr/pages/html/index.tsx
```diff
@@ -37,7 +39,7 @@ export default class Html extends React.Component {
           <Link href="/">
             Back to home
           </Link>
-          <MatchCommandClassComponent commands={expressionToMatch} />
+          <MatchCommandClassComponent commands={expressionToMatch} successMessages={matchSuccess}/>
```
iampava, line 42: Instead of having 2 separate props, let's combine them into an Array of Objects, each with two properties. It's easier to manage them this way.

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
...
+                {' '}
+                {message.time}
+              </p>
+            ))}
+          </div>
+          <form
+            className={styles.form}
+            onSubmit={this.handleSubmit}
+          >
+            <input
```
iampava, line 121: Let's add `spellCheck="false"` as an attribute. Since this commands may not be valid english, we don't want the red lines underneath.

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
+import React from 'react';
+import {uuid} from '../../services/Utils';
```
iampava, line 2: Make sure you fix all ESLint issues. For example, here there should be a space, like so `{ uuid }`. 

You should see some red lines underneath the code with problems. If you don't maybe it's not correctly configured
image

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
...
+    } else {
+      isMatch = commands[currentIndex] === gitCommand.trim().toLowerCase();
+    }
+
+    const date = new Date();
+
+    const commandMessage: Message = {
+      text: gitCommand,
+      type: MessageType.COMMAND,
+      time: date.toLocaleTimeString(),
```
iampava, line 62: Let's not convert it to string here, but in the render function. Maybe later we want to do some checks/verificatins regarding those timestamps, and it will be easier if they are numbers.

So let's use: `time: Date.now()`

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
...
+      id: uuid()
+    };
+  
+    const successMessage : Message = {
```
iampava, line 66: It's a waste to create both the success and error message. We already know based on the `isMatch` variable in which situation we are.

So use just one variable, and assign different values depending on the case.

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
...
+    return (
+      <>
+        <div className={styles.commands}>
+          <div
+            className={styles.commands__cli}
+          >
+            {messages.map((message) => (
+              <p key={message.id} className={styles[message.type]}>
+                {message.text}
+                {' '}
```
iampava, line 112: We can delete this whitespace, we don't need it.

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
...
+  };
+
+  render() {
+    const { messages, btnMessage } = this.state;
+
+    return (
+      <>
+        <div className={styles.commands}>
+          <div
+            className={styles.commands__cli}
```
iampava, line 107: I think this should be on the same line, no?

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
...
+      time: date.toLocaleTimeString(),
+      id: uuid()
+    };
+
+    if (!isMatch) {
+      this.setState(({ messages }) => ({
+        messages: [...messages, commandMessage, errorMessage],
+        btnMessage: 'Next',
+      }));
+    } else if (currentIndex === commands.length - 1) {
```
iampava, line 85: Here should be `commands.length - 2`. We want the "Done" text to appear before the last command.

Also, make sure the text inside the button is correct for all cases, including when having just one item in the Array. In that case, it should say `Done` from the beginning.

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,135 @@
...
+      this.setState(({ messages }) => ({
+        messages: [...messages, commandMessage, successMessage],
+        currentIndex: 0,
+        btnMessage: 'Done',
```
iampava, line 89: If we're done, let's disable the input so we can't submit the form any more.

iampava (changes requested):

frontend-ssr/components/cli_git/GitCli_CC.tsx
andreeatoma: Added the ESLint extension in VSCode.

frontend-ssr/components/cli_git/GitCli_CC.tsx
andreeatoma: I don't like how it looks if i delete the whitespace. 
image
Can I use something else?

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,145 @@
...
+}
+
+class matchCommandMessageClassComponent extends React.Component<Props, State, Message> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      currentIndex: 0,
+      messages: [],
+      btnMessage: 'Next',
```
iampava, line 38: Here we should treat the case in which we have just one item in the Array. Meaning it should be like this:

`btnMessage: props.commands.length === 1 ? 'Done' : 'Next'`

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,145 @@
...
+      matchCommandMessage = {
+        text: successMessages[currentIndex],
+        type: MessageType.SUCCESS,
+        time: Date.now(),
+        id: uuid(),
+      };
+    }
+
+    if (!isMatch) {
+      this.setState(({ messages }) => ({
```
iampava, line 85: A little bit of duplicate code between those if statements. I think we can simplify them a lot.

First thing I notice is that no matter the case, we always insert the command and the command message into the `messages` Array. So let's move that state change outside the if statements.

```
this.setState(({ messages }) => ({
   messages: [...messages, commandMessage, matchCommandMessage],
   currentIndex: isMatch ? currentIndex + 1: currentIndex
}));
```

After that, I think we only need one `if statement`, to check if we need to change the button text

```
if (isMatch && currentIndex + 2 === commands.length) {
  // One before the end
  this.setState({
    btnMessage: 'Done'
  })
}
```

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,145 @@
...
+interface Props {
+  commands: RegExp[] | string[],
+  successMessages: string[],
+}
+
+interface State {
+  currentIndex: number,
+  messages: Message[],
+  btnMessage: string,
+  loading: boolean,
```
iampava, line 15: We don't need a separate value to see if we need to disable the form. We can do this based on `currentIndex`. I put a comment in the render method regarding this.

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,145 @@
...
+              type="text"
+              name="gitCommand"
+              spellCheck="false"
+              required
+            />
+            <span className={styles.blinking__cursor} />
+            <button
+              className="btn btn--blue"
+              type="submit"
+              disabled={loading}
```
iampava, line 135: As mentioned before, we don't need another variable for this. We can simply check if `currentIndex === commands.length`. This means we've reached the end. 

Also, forgot to tell you in the last review. Add a `readOnly` attribute to the input when we've finished everything, so that you cannot write there.

frontend-ssr/pages/html/index.tsx
```diff
@@ -37,7 +41,10 @@ export default class Html extends React.Component {
           <Link href="/">
             Back to home
           </Link>
-          <MatchCommandClassComponent commands={expressionToMatch} />
+          <MatchCommandClassComponent
+            commands={expressionToMatch.map((item) => item.command)}
```
iampava, line 45: Sorry, i don't think I was explicit enough. Let's modify the props as well, instead of 2 let's have just one, the array of objects.

Same logic as above, better for them to pe "packed" together

iampava (changes requested):

frontend-ssr/components/cli_git/GitCli_CC.tsx
```diff
@@ -0,0 +1,141 @@
...
+  id: string,
+}
+
+enum MessageType {
+  COMMAND = 'command',
+  ERROR = 'error',
+  SUCCESS = 'success'
+}
+
+class matchCommandMessageClassComponent extends React.Component<Props, State, Commands> {
```
iampava, line 34: Class components should be `PascalCase` meaning `MatchCommandMessageClassComponent`. Also, why the 3rd param, shouldn't it be just `<Props,State>`?

iampava (changes requested):

PR #27, opened 2020-09-10
<TableOfContents> component

andreeatoma:


frontend-ssr/components/custom_menu/Menu.module.scss
```diff
@@ -0,0 +1,52 @@
...
+        &.active,
+        &:hover {
+            color: color(purple);
+            &::after {
+                background-color: color(purple);
+                transition: all .3s ease-in;
+            }
+        }
+        &:last-child {
+            &::before {
```
iampava, line 47: In situations like this there's no need for additional new-line. We can do

```
&:last-child::before {
  // ...
}
```

frontend-ssr/components/custom_menu/Menu.module.scss
```diff
@@ -0,0 +1,52 @@
...
+        background: none;
+        border: none;
+        outline: none;
+        color: color(black);
+        font-weight: bolder;
+        cursor: pointer;
+        padding-left: 2.5em;
+        min-height: 60px;
+        position: relative;
+        &::before {
```
iampava, line 16: I would expect `before & after` to be reversed. The `line` to be `after` since it comes after the button, and the Dot to be `before`.

iampava (changes requested):

iampava:
A usability concern I found is that buttons are too tall, and you can click in between them. This shouldn't be possible
2020-09-11_09h58_19

PR #28, opened 2020-09-12
<TableOfContents /> component

andreeatoma:


frontend-ssr/components/tableOfContents/TableOfContents.module.scss
```diff
@@ -0,0 +1,48 @@
+@import '../../styles/variables';
+@import '../../styles/buttons';
```
iampava, line 2: Why are we importing this the buttons style here?

frontend-ssr/components/tableOfContents/TableOfContents.module.scss
```diff
@@ -0,0 +1,48 @@
...
+            content: '';
+            position: absolute;
+            top: 42%;
+            left: 0.8em;
+            background-color: color(grey);
+            width: 0.13em;
+            height: 50%;
+        }
+        &.active,
+        &:hover {
```
iampava, line 35: There's a small usability problem if these get highlighted when hovering the `<li>`. This gives me the impression I can click them, however, I can only do it if I click on the `<button>`.

So, the expected behavior is that if the dot gets highlighted, I can also click it and scroll.

2020-09-12_15h19_24

frontend-ssr/components/tableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,65 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeTitleChapter: null,
+    };
+  }
+
+    handleScrollItem = (id) => {
+      let itemToScroll = document.getElementById(id);
```
iampava, line 27: I just discovered an even easier way to do this, from a friend, including the scroll smooth.  Sorry, I didn't know about this from the beginning. :D 

Instead of using `scrollIntoView` we can change the button to an ```<a href={`#${item.id}`} />```. and then also add in `index.scss` the rule `scroll-behavior: smooth;` on `html` - https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior.

We also need to keep the `onClick` listener but just to set the active item so we know which dot to highlight.

iampava: Let's also keep the verification regarding `itemToScroll` so we know if there's something missing in the config.

frontend-ssr/components/tableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,65 @@
...
+      const { chapters } = this.props;
+      const { activeTitleChapter } = this.state;
+      return (
+        <ul className={styles.tableOfContents}>
+          {chapters.map((item) => (
+            <li
+              key={item.id}
+              className={activeTitleChapter === item.id ? styles.active : ''}
+            >
+              <button
```
iampava, line 52: As mentioned above, change this to an `<a>`. But we can keep the classes which we need.

frontend-ssr/components/tableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,65 @@
...
+    title: string,
+    id: string,
+}
+
+class TableOfContents extends React.Component<Props, State> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeTitleChapter: null,
```
iampava, line 22: This is still named wrong. `activeTitleChapter` sounds to me like  it's holding some type of title, but instead it;s holding ids. Let's rename to `activeChapterId`

frontend-ssr/components/tableOfContents/TableOfContents.module.scss
```diff
@@ -0,0 +1,48 @@
+@import '../../styles/variables';
+@import '../../styles/buttons';
+.tableOfContents {
+    li {
```
iampava, line 4: `ul` elements come with a default padding. Let's remove it. I think the best way is to assume we always remove this for all `ul/ol` elements, so you can add this style in `index.scss` for now.

iampava (changes requested):

frontend-ssr/components/tableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,96 @@
...
+  componentWillUnmount = () => {
+    this.observer.disconnect();
+  }
+
+  handleScrollItem = (id) => {
+    let itemToScroll = document.getElementById(id);
+    if (itemToScroll === null) {
+      console.warn('TableOfContents: expected to find element with id="id" to scroll to');
+    }
+    this.observer.unobserve(itemToScroll);
```
iampava, line 64: Here we need to unobserve all elements. It's not enough if we do just for this one, unfortunately.

Also, please add an explainer regarding what's the purpose of this, what problem we're trying to solve., for future reference.

frontend-ssr/components/tableOfContents/TableOfContents.module.scss
```diff
@@ -0,0 +1,50 @@
...
+            &.active,
+            &:hover {
+                color: color(green);
+                &::before {
+                    background-color: color(green);
+                    transition: background-color .3s ease-in;
+                }
+            }
+        }
+        &:last-child {
```
iampava, line 42: If we have just one child, no need to nest. We can just do

```
&:last-child a::after {
  ...
}
```

frontend-ssr/components/tableOfContents/TableOfContents.tsx
```diff
@@ -0,0 +1,96 @@
...
+        if (entryId !== null && entry.intersectionRatio > 0.5) {
+          // remove old active class
+          document.querySelector(`.${styles.active}`).classList.remove(`${styles.active}`);
+          // find matching link & add active class
+          document.querySelector(`[href="#${entry.target.id}"]`).classList.add(`${styles.active}`);
+        }
+      });
+    };
+
+    const options = {
```
iampava, line 41: After you refactor the code from the `changeNav` function, let's add one small improvement. We need to change the URL when we have a match, otherwise if we refresh the page it will lead us to the old chapter.

To do this you're gonna have to use the NextJS router object, specifically it's `.replace` function. - https://nextjs.org/docs/api-reference/next/router

Be careful, here we have a Class Component so you cannot use the hook. You're gonna have to wrap it with `withRouter`, so make sure you read the docs.

iampava (changes requested):
Overall looks sweet, just how I imagined it. We can do some small improvement until we merge it though.

PR #32, opened 2020-09-25
Exercise Submissions functionality

andreeatoma:
CRUD exercise

store and data modelling


frontend-ssr/redux/exercise-submissions/submissions.reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+): SubmissionState => {
+  switch (action.type) {
+    case 'GET_SUBMISSIONS':
+      return {
+        ...state.submissions,
+        submissions: action.payload,
+      };
+    case 'ADD_SUBMISSION':
+      return {
+        submissions: [...state.submissions, action.payload],
```
iampava, line 18: Let's add a parameter here which is the index where to insert the submission. It will come in handy

iampava: Also, let's merge ADD & Get. There's no point in having both of them. We can just keep the `/ADD` one and make it in such a way that we can insert multiple values at once.

frontend-ssr/redux/exercise-submissions/submissions.reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+      };
+    case 'REMOVE_SUBMISSION':
+      return {
+        submissions: state.submissions.filter((submission) => submission.id !== action.payload),
+      };
+    case 'EDIT_SUBMISSION':
+      return {
```
iampava, line 25: I don't think we need to keep this one in store.

frontend-ssr/redux/exercise-submissions/submissions.reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+          if (submission.id === action.id) {
+            return {
+              ...submission,
+              username: action.payload.newUsername,
+              avatarSrc: action.payload.newavatarSrc,
+              chapter: action.payload.newChapter,
+              type: action.payload.newType,
+              editing: !submission.editing,
+            };
+          } return submission;
```
iampava, line 42: Eslint should mark this and force an `Enter`, no?

frontend-ssr/redux/exercise-submissions/types.ts
```diff
@@ -0,0 +1,21 @@
...
+    chapter: ChapterType,
+    time: number,
+    type: SubmissionType,
+    editing?: boolean,
+}
+export interface SubmissionState {
+    submissions: Submission[],
+}
+export enum ChapterType {
+    HTML = 'html',
```
iampava, line 14: This enum should be in a generic `constants.ts` file. I have a feeling we'll use it in multiple parts of the app.

frontend-ssr/redux/exercise-submissions/submissions.actions.ts
```diff
@@ -0,0 +1,27 @@
...
+});
+export const removeSubmission = (id) => ({
+  type: REMOVE_SUBMISSION,
+  payload: id,
+});
+export const editSubmission = (id) => ({
+  type: EDIT_SUBMISSION,
+  payload: id,
+});
+export const updateSubmissions = (newSubmission, id) => ({
```
iampava, line 23: Let's put the id first. This should be a general pattern throughout the app.

iampava: Also, this should be in singular: `updateSubmission(...)`

frontend-ssr/redux/exercise-submissions/submissions.reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+          ? { ...submission, editing: !submission.editing }
+          : submission)),
+      };
+    case 'UPDATE_SUBMISSION':
+      return {
+        submissions: state.submissions.map((submission) => {
+          if (submission.id === action.id) {
+            return {
+              ...submission,
+              username: action.payload.newUsername,
```
iampava, line 36: The properties of `action.payload` should be the same as the submission. It's easier if we mantain the same interface, instead of having both `username`, `newUsername`, etc... Also, you shouldn't check for each property in particular. Imagine if our data structure grows and we'll have 10+ props :( 

So let's do

```
return { 
  ...submission,
  ...action.payload.submission
}
```

frontend-ssr/redux/exercise-submissions/submissions.reducer.ts
```diff
@@ -0,0 +1,49 @@
...
+    case 'EDIT_SUBMISSION':
+      return {
+        submissions: state.submissions.map((submission) => ((submission.id === action.payload)
+          ? { ...submission, editing: !submission.editing }
+          : submission)),
+      };
+    case 'UPDATE_SUBMISSION':
+      return {
+        submissions: state.submissions.map((submission) => {
+          if (submission.id === action.id) {
```
iampava, line 33: There is no such thing as `action.id`. Actions only have `type` and `payload`.

frontend-ssr/redux/exercise-submissions/submissions.actions.ts
```diff
@@ -0,0 +1,27 @@
...
+  type: UPDATE_SUBMISSION,
+  payload: newSubmission,
+  id,
```
iampava, line 26: No no no. Actions only have `type` and `payload`.

iampava (changes requested):

frontend-ssr/package-lock.json
```diff
@@ -0,0 +1,8461 @@
+{
```
iampava, line 1: I think you ran `npm install` here. 

We're using `yarn` so there should be no `package-lock.json` file.

frontend-ssr/redux/constants.ts
```diff
@@ -0,0 +1,7 @@
+enum ChapterType {
```
iampava, line 1: Lets move this file into `/types` file you created below.

frontend-ssr/redux/exercise-submissions/exercise-submissions.actions.ts
```diff
@@ -0,0 +1,30 @@
+import { Submission } from './types';
+
+const ADD_SUBMISSION = 'exercise-submissions/ADD';
```
iampava, line 3: Let's simplify the namings to just `ADD`, `REMOVE`, `UPDATE`.

frontend-ssr/redux/exercise-submissions/exercise-submissions.actions.ts
```diff
@@ -0,0 +1,30 @@
...
+const REMOVE_SUBMISSION = 'exercise-submissions/REMOVE';
+const UPDATE_SUBMISSION = 'exercise-submissions/UPDATE';
+
+export const EXERCISE_SUBMISSIONS = {
+  ADD_SUBMISSION,
+  REMOVE_SUBMISSION,
+  UPDATE_SUBMISSION,
+};
+
+export const addSubmission = (submission: Submission, index: number) => ({
```
iampava, line 13: The types are wrong. If we can either send a submission or an array, we should have 

```
(data : Submission | Submissions[], index: number)
```

Also, let's add a default value to index.

frontend-ssr/redux/exercise-submissions/exercise-submissions.actions.ts
```diff
@@ -0,0 +1,30 @@
...
+  type: ADD_SUBMISSION,
+  payload: {
+    submission,
+    index,
+  },
+});
+export const removeSubmission = (id: string) => ({
```
iampava, line 20: Same here. If we support single or multiple, types should be `string | string[]`.

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,54 @@
+import { SubmissionState } from './types';
+import { EXERCISE_SUBMISSIONS } from './exercise-submissions.actions';
+
+const initialState: SubmissionState = {
+  submissions: [],
+};
+export const submissionReducer = (state = initialState, action: { type: string; payload: any;}) => {
```
iampava, line 7: Please add types for the return object, meaning  `: SubmissionState`. This way we would catch errors for bad types.

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,54 @@
...
+  switch (action.type) {
+    case EXERCISE_SUBMISSIONS.ADD_SUBMISSION: {
+      const { index, submission } = action.payload;
+      const newArray = [...state.submissions];
+      newArray.splice(index, 0, submission);
+      if (Array.isArray(submission)) {
+        submission.map((item) => ({
```
iampava, line 14: Did you test this. Doesn't seem like the result of this map is being put anywhere. Looks a little wrong as well

Also, I think we can simply to something like:

```
[...state.submissions.slice(0, index),  ...submission, ...state.submissions.slice(index)]
```

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,54 @@
...
+        }));
+      }
+      return {
+        newArray,
+        submissions: [...state.submissions, submission],
+      };
+    }
+    case EXERCISE_SUBMISSIONS.REMOVE_SUBMISSION: {
+      const { id } = action.payload;
+      if (Array.isArray(id)) {
```
iampava, line 26: Did you test this function when sending an Array It doesn't seem to be doing anything...

frontend-ssr/components/exercise-submissions/ExerciseSubmissionsList.tsx
```diff
@@ -0,0 +1,50 @@
...
+            )
+            }
+        </div>
+      ))}
+    </div>
+  );
+}
+
+const mapStateToProps = (state: any) => ({
+  submissions: state,
```
iampava, line 47: We don't wanna get the entire state here. Just the part we're interested in.

frontend-ssr/redux/root.reducer.ts
```diff
@@ -1,8 +1,8 @@
 import { combineReducers } from 'redux';
-import { dummy } from './dummy/dummy-reducer';
+import { submissionReducer } from './exercise-submissions/exercise-submissions.reducer';
 
 const rootReducer = combineReducers({
-  dummy,
+  submissionReducer,
```
iampava, line 5: The state is a plain object so the property names should not be related to redux.

So here, let's change to

```
import {submissionReducer as submissions} from '...';

combineReducers({
  submissions
})
```

iampava (changes requested):

iampava:
Also 

1. please change commit name to use the pattern described in `CONTRIBUTING.md`.
2. change the name of this PR to `Exercise Submissions functionality`. We're gonna merge when everything is ready so no need for "first steps" in name

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,54 @@
...
+const initialState: SubmissionState = {
+  submissions: [],
+};
+export const submissionReducer = (state = initialState, action: { type: string; payload: any;})
+: SubmissionState => {
+  switch (action.type) {
+    case EXERCISE_SUBMISSIONS.ADD: {
+      let { index, submission } = action.payload;
+      if (Array.isArray(submission)) {
+        submission = [...submission];
```
iampava, line 13: It's already an Array, no need for this case. We only care about the one in which it's not an Array.

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+      return {
+        submissions: [...state.submissions.slice(0, index),
+          ...submission,
+          ...state.submissions.slice(index)],
+      };
+    }
+    case EXERCISE_SUBMISSIONS.REMOVE: {
+      let { id } = action.payload;
+      if (Array.isArray(id)) {
```
iampava, line 25: We can apply the same approach as above, if it's not an Array treat it as such. Then we will have just one return statement.

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,54 @@
...
+      return {
+        submissions: state.submissions.map((submission) => {
+          const { id, newSubmission } = action.payload;
+          if (submission.id === id) {
+            return {
+              ...submission,
+              ...newSubmission,
+            };
+          }
+          return {
```
iampava, line 45: This is buggy. Should be `return submission`.

iampava (changes requested):
Only a few minor things remaining. 

Please modify the commit name as well, use CamelCase like so `[ExerciseSubmissions] Basic data modelling and store`

No need to change the PR description, that one is fine as it is.

frontend-ssr/components/exercise-submissions/ExerciseSubmissionsPage.tsx
```diff
@@ -33,7 +33,6 @@ function ExerciseSubmissionPage(props: { submissions, dispatch }) {
       <br />
       <ul>
         {submissions.submissions.map((submission) => (
-          console.log(submissions),
```
iampava: This commit should be squashed with the second one so we have just two in total.

frontend-ssr/redux/exercise-submissions/exercise-submissions.actions.ts
```diff
@@ -30,3 +36,21 @@ export const updateSubmission = (id: string, newSubmission: Submission) => ({
...
+});
+
+export const searchSubmissions = (keyword: string) => ({
+  type: SEARCH,
+  payload: {
+    keyword,
+  },
+});
+
+export const nextPage = () => ({
```
iampava, line 54: No need for `nextPage` action or action creator for now. We'll need it in step 3 when we add some async work

frontend-ssr/redux/exercise-submissions/types.ts
```diff
@@ -10,6 +10,8 @@ export interface Submission {
 }
 export interface SubmissionState {
     submissions: Submission[],
+    currentPage: number,
+    countPerPage: number,
```
iampava, line 14: Although this is generally a good idea, let's keep it as simple as possible for now. No need for a configurable `countPerPage`

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -1,8 +1,12 @@
 import { SubmissionState } from './types';
 import { EXERCISE_SUBMISSIONS } from './exercise-submissions.actions';
+import getMockSubmissions from '../../services/Helpers';
+import MOCK_SUBMISSIONS from '../../pages/api/exercise-submissions/mock';
 
 const initialState: SubmissionState = {
-  submissions: [],
+  submissions: MOCK_SUBMISSIONS,
```
iampava, line 7: We should add another field here keeping track of the search query. Let's name it `search`. 

Also, we can rename `currentPage` to `page`, I think it's obvious what's going on.

frontend-ssr/redux/exercise-submissions/exercise-submissions.actions.ts
```diff
@@ -30,3 +36,21 @@ export const updateSubmission = (id: string, newSubmission: Submission) => ({
     newSubmission,
   },
 });
+
+export const loadSubmissions = (n: number) => ({
+  type: LOAD,
```
iampava, line 41: This action creator should already receive the new submissions as payload. In other words, all the mocking should be done beforehand, in `<ExerciseSubmissionsPage />`.

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -40,6 +47,32 @@ export const submissionReducer = (state = initialState, action: { type: string;
           return submission;
         }),
       };
+    case EXERCISE_SUBMISSIONS.LOAD: {
+      const { n } = action.payload;
+      const submissions = getMockSubmissions(n);
+      return {
+        ...state,
+        submissions,
```
iampava, line 55: Here we should append the new submissions to the list, not replace them altogether. 

Also, please increae the page numbe as well.

frontend-ssr/components/exercise-submissions/ExerciseSubmissionsPage.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { submissions, dispatch } = props;
+
+  const onSearchSubmission = (e) => {
+    e.preventDefault();
+    let keyword = e.target.searchKeyword.value;
+    dispatch(searchSubmissions(keyword));
+  };
+
+  function onNextPage() {
+    dispatch(loadSubmissions(submissions.countPerPage));
```
iampava, line 18: No need for 2 separate actions. Let's call just load with the already mocked "new" submissions.

iampava (changes requested):
Apart from the mentioned changes, I'd like you to also work on UI a little bit so it  works like this:

1. Enter page -> see some mocke submissions
2. Search by `a`. The filtered submissions appear. 
3. Click next page -> mocked submissions that match the query are added
4. Delete everything from search and click enter, the initial submissions appear

frontend-ssr/services/Constants.ts
```diff
@@ -4,4 +4,4 @@ enum ChapterType {
   JS = 'js',
 }
 
-export { ChapterType as default };
+export default ChapterType;
```
iampava, line 7: No default export in constants. We'll add many more things here so ;let's export them separately.

iampava: Also, since you're at it please rename to lowercase - `constants.ts` :D

frontend-ssr/services/ExerciseSubmissions.service.ts
```diff
@@ -0,0 +1,18 @@
+import MOCK_SUBMISSIONS from '../pages/api/exercise-submissions/mock';
+
+const LOAD_DURATION = 1000;
+
+export default async function getSubmissions(page: number, query: string) {
+  let promise = new Promise((resolve, reject) => {
```
iampava, line 6: You mixed `async/await` with Promises which is a big no-no. One or the other.

If you wanted to go with an async/await approach it would be something like:

```
async function getSubmissons(...){
  await wait(1000);
  return MOCK_SUBMISSIONS; 
}

function wait(mili) { 
  // Create another helper function which returns a promise that resolves after `mili` milliseconds
  return new Promise((resolve) => setTimeout(resolve, mili));
}


```

iampava (changes requested):

iampava (changes requested):

iampava:
A new PR has been opened into a different upstream branch. Closing this one and moving conversation there

PR #34, opened 2020-10-12
Exercise Submissions functionality

andreeatoma:


frontend-ssr/services/Helpers.ts
```diff
@@ -0,0 +1,21 @@
+import { uuid } from './Utils';
```
iampava, line 1: I think we can remove this one since we have integrated with the BackEnd.

frontend-ssr/services/ExerciseSubmissions.service.ts
```diff
@@ -0,0 +1,8 @@
+import HttpService from './Http.service';
+
+export default async function getSubmissions(page: number, query: string) {
+  const responseSubmissions = await HttpService.get(`${process.env.ENDPOINT}/exercises/submissions?search=al`)
```
iampava, line 4: We should use the params when creating the GET URL. Also I think we should have default values for page as well as query (0 and `''`)

frontend-ssr/services/ExerciseSubmissions.service.ts
```diff
@@ -0,0 +1,8 @@
...
+    .then((response) => response.json())
+    .catch((err) => alert(err));
```
iampava, line 6: We shouldn't handle the error here but in the react component. Also, no point for async-await or `.then` here.

I think the HttpServic already handles converting to `json` so in this function you should just return the call to the service, no async-await needed.

PS: let's add some types as well, that would be amazing:

`: Promise<Submission[]>`

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,78 @@
+import { SubmissionState } from './types';
+import { EXERCISE_SUBMISSIONS } from './exercise-submissions.actions';
+import MOCK_SUBMISSIONS from '~/pages/api/exercise-submissions/mock';
```
iampava, line 3: Shouldn't need MOCK_SUBMISSIONS any more.

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,78 @@
...
+              ...newSubmission,
+            };
+          }
+          return submission;
+        }),
+      };
+    case EXERCISE_SUBMISSIONS.LOAD: {
+      const { newSubmissions } = action.payload;
+      if (state.submissions !== undefined) {
+        return {
```
iampava, line 52: A little duplicate code here. Let's use the ternary operator

```
return {
  ...state,
  submissions: state.submissions ? [ ... ] : newSubmissions,
  page: state.page,
}
```
```

File: frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts [hunk 6]
```diff
@@ -0,0 +1,78 @@
+import { SubmissionState } from './types';
+import { EXERCISE_SUBMISSIONS } from './exercise-submissions.actions';
+import MOCK_SUBMISSIONS from '~/pages/api/exercise-submissions/mock';
+
+const initialState: SubmissionState = {
+  submissions: undefined,
+  search: '',
+  page: 1,
```
iampava, line 8: Let's use 0-based index for page number. Otherwise it may be confusing, since almost all things in JS start from index 0.

frontend-ssr/pages/api/exercise-submissions/mock.ts
```diff
@@ -0,0 +1,23 @@
+import { uuid } from '../../../services/Utils';
```
iampava, line 1: We can delete this, right?

iampava (changes requested):

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -0,0 +1,72 @@
+import { SubmissionState } from './types';
+import { EXERCISE_SUBMISSIONS } from './exercise-submissions.actions';
+
+const initialState: SubmissionState = {
+  submissions: undefined,
+  search: '',
+  page: 1,
```
iampava, line 7: I see you changed the page number at the botttom, but you should have changed it here since 0 is the starting one.

frontend-ssr/services/ExerciseSubmissions.service.ts
```diff
@@ -0,0 +1,11 @@
+import { Submission } from '~/redux/exercise-submissions/types';
+import HttpService from './Http.service';
+
+// FIXME ---> we need to do a little more research to find a better solution
+
+class ExerciseService {
+  static getSubmissions(page = 0, query = '') : Promise<Submission[]> {
+    return HttpService.get(`${process.env.ENDPOINT}/exercises/submissions?search=al${page}&${query}`) as unknown as Promise<Submission[]>;
```
iampava, line 8: Probably in a hurry but you left `?search-al` in the request string. I put it there just as an example. 

So it should be `/submissions?page=${page}&query=${query}` - this is how query params work

iampava (changes requested):

PR #35, opened 2020-10-14
[ExerciseSubmissions] Integrate UI and Redux

andreeatoma:


frontend-ssr/components/exercise-submissions/SubmissionSkeleton.module.scss
```diff
@@ -0,0 +1,42 @@
+@import '../../styles/variables';
```
iampava, line 1: All this code is duplicated from the other SCSS file. This should have "ringed" a warning bell and make you question if this is the right approach. Let's merge the two style files and remove the duplicated code.

frontend-ssr/components/exercise-submissions/SubmissionSkeleton.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import styles from './SubmissionSkeleton.module.scss';
+import { Submission, SubmissionType } from '~/redux/exercise-submissions/types';
+
+interface Props {
+  submission: Submission,
+}
+export default function SubmissionSkeleton({ submission } : Props) {
```
iampava, line 8: It seems the markup of this component is pretty much the same as the main component. This shouldn't be the case at all. 

A skeleton component should be much much MUCH simpler: just a couple of placeholder elements without any real content. So this should be refactored and have just empty elements without any props or content.

frontend-ssr/components/exercise-submissions/Submission.module.scss
```diff
@@ -0,0 +1,53 @@
...
+    .card_date {
+        letter-spacing: 0.01em;
+    }
+    .card_status_pending,
+    .card_status_assigned {
+        min-width: 10em;
+        @extend %gradientBackground,
+        %greenGradientBackground;
+    }
+    .card_status_pending {
```
iampava, line 35: 2 separate rules for the same class name: `card_status_pending`?

Also, all class names should be `kebab-case`

frontend-ssr/styles/_animations.scss
```diff
@@ -1,11 +1,22 @@
...
+        opacity: 1;
+    }
+    100% {
+        opacity: 0;
+    }
 }
+
+@keyframes loading {
+    0% {
+        background-position: -500px 0;
```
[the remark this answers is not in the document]
iampava: Also let's name this `skeletonOpacity` or something like this, because we'll be reusing it for all skeletons.

iampava (changes requested):

frontend-ssr/components/exercise-submissions/Submission.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Submission, SubmissionType } from '../../redux/exercise-submissions/types';
+
+import styles from './Submission.module.scss';
+
+interface Props {
+    submission: Submission,
+}
+function ExSubmission({ submission } : Props) {
```
iampava, line 9: Component should be named `ExerciseSubmission`. It's important to keep the same naming everywhere, his way when we search for this string we find all the files in which it's used.

Also for the skeleton, it should be `ExerciseSubmissionSkeleton`

iampava (changes requested):

frontend-ssr/styles/_animations.scss
```diff
@@ -1,11 +1,22 @@
...
-  0% {
-    opacity: 0;
-  }
-  50% {
-    opacity: 1;
-  }
-  100% {
-    opacity: 0;
-  }
+    0% {
```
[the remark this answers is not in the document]
andreeatoma: Thank you Alex. I have checked my Workspace Settings and the configuration seems to be ok.
image
Hope that now we are "on the same page" because I have set up those commands too.

frontend-ssr/components/exercise-submissions/ExerciseSubmissionSkeleton.tsx
```diff
@@ -0,0 +1,32 @@
...
+            />
+            <span />
+          </a>
+          <span className={`${styles['card-el']} ${styles['card-chapter']} rounded-md bg-white text-xs`} />
+        </div>
+        <span
+          className={`${styles['card-el']} ${styles['card-status-assigned']} rounded-md bg-white text-xs`}
+        />
+      </div>
+      <time />
```
iampava, line 29: We can use the `rounded-md` class here and get rid of another CSS rule.

.vscode/settings.json
```diff
@@ -7,5 +7,10 @@
     "typescriptreact"
   ],
   "eslint.workingDirectories": ["./frontend-ssr"],
-  "editor.tabSize": 2
+  "editor.tabSize": 2,
+  "emmet.excludeLanguages": [
```
iampava, line 11: Hmm, not sure we need those? Let's have a sync today and figure this indentation problem together :D

frontend-ssr/components/exercise-submissions/ExerciseSubmission.tsx
```diff
@@ -0,0 +1,62 @@
...
+              className={`${styles['card-el']} ${styles['card-status-pending']} text-bold rounded-md bg-grey text-center text-xs`}
+            >
+              <option value="pending">{SubmissionType.PENDING}</option>
+              <option value="assignTo">assign-to-you</option>
+            </select>
+          )
+        }
+
+      </div>
+      <time>{new Date(submission.time).toLocaleString()}</time>
```
iampava, line 57: Let's parse the time, and show it like this depending on when it was submitted:

1. `Just now` - if the submission was under one minute ago
2. `X minutes ago` or `Y hours ago`
3. `Yesterday`
3. `X days ago` - up to 3 days
4. `10 Sept 2020` if it as more than 3 days ago

iampava (changes requested):

frontend-ssr/components/exercise-submissions/ExerciseSubmissionSkeleton.tsx
```diff
@@ -0,0 +1,32 @@
...
+import styles from './ExerciseSubmission.module.scss';
+
+export default function SubmissionSkeleton() {
+  return (
+    <div
+      className={`${styles.card} ${styles['skeleton-screens']} text-white rounded-md bg-grey`}
+    >
+      <div className={`${styles['card-top']} d-flex align-items-center justify-content-between`}>
+        <div className={`${styles['card-info']} d-flex align-items-center justify-content-between`}>
+          <a
```
andreeatoma, line 11: This one is already an <a>, right? I don't really understand the problem here.

iampava: Oups, my bad. I meant replace this `<a>`. No point in a having a link pointing to nowhere....remember we don't know the content beforehand.

frontend-ssr/services/Utils.ts
```diff
@@ -1 +1,26 @@
 export { v4 as uuid } from 'uuid';
+
+export const timeAgo = (date) => {
```
iampava, line 3: What type is `date` and the return type of the function?

frontend-ssr/services/Utils.ts
```diff
@@ -1 +1,26 @@
...
+  let ms = (new Date()).getTime() - date.getTime();
```
iampava, line 4: To get the current number of milliseconds we can use directly `Date.now()`

frontend-ssr/services/Utils.ts
```diff
@@ -1 +1,26 @@
...
+  let seconds = Math.floor(ms / 1000);
+  let minutes = Math.floor(seconds / 60);
+  let hours = Math.floor(minutes / 60);
+  let days = Math.floor(hours / 24);
+
+  if (seconds < 60) {
+    return 'Just now';
+  } if (minutes < 60) {
```
iampava, line 12: Did you format this code? This if should be on another line.

frontend-ssr/services/Utils.ts
```diff
@@ -1 +1,26 @@
...
+    return minutes === 1
+      ? `${minutes} minute ago`
+      : `${minutes} minutes ago`;
+  } if (hours < 24) {
```
iampava, line 16: Actually all if's should be on new lines. It makes everything easier to read:

```
if () {
  ...
}

if () {
  ... 
}

```

frontend-ssr/services/Utils.ts
```diff
@@ -1 +1,26 @@
...
+    return hours === 1
+      ? `${hours} hour ago`
+      : `${hours} hours ago`;
+  } if (days < 24 * 3) {
```
iampava, line 20: Let's add another verification for "Yesterday"

frontend-ssr/services/Utils.ts
```diff
@@ -1 +1,26 @@
...
+    return `${days} days ago`;
+  } if (days > 24 * 3) {
+    return date;
```
iampava, line 23: Shouldn't we parse this date somehow? The format we're looking for is -> `20 September 2020`

frontend-ssr/services/Utils.ts
```diff
@@ -1 +1,26 @@
...
+  }
+  return false;
```
iampava, line 25: Uhm, we should only return strings from this function. Don't quite understand why we need the `false` here.

frontend-ssr/services/Utils.ts
iampava: Also, let's a comment documenting the different return types. Explain a little bit what the function does and then the different formats it will return, it can be with bullet points.

iampava (changes requested):

PR #36, opened 2020-10-20
[ExerciseSubmissions] Basic BE  & Redux integration

andreeatoma:


frontend-ssr/pages/submissions.module.scss
```diff
@@ -0,0 +1,31 @@
+@import "../styles/variables";
+.page-submissions {
+  width: $width;
+  max-width: $max-width;
+  form {
```
iampava, line 5: Uhmm...there's no form in the rendered elements? 😅

frontend-ssr/pages/submissions.module.scss
```diff
@@ -0,0 +1,31 @@
...
+      &::-moz-focus-inner {
+        border: 0;
+      }
+    }
+    button {
+      border-bottom-left-radius: 0;
+      border-top-left-radius: 0;
+    }
+  }
+  .cardWrapper {
```
iampava, line 26: This should be in plural. Also it should be in kebab-case.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,109 @@
...
+    super(props);
+    this.state = {
+      loading: true,
+    };
+  }
+
+  componentDidMount() {
+    let { submissions } = this.props;
+    const state = submissions;
+    function shouldFetchSubmissions() {
```
iampava, line 35: This entire thing seems overly complicated. You have another inner function declared here, which is usually not a good practice, then you also call the store directly and get it's state, but don't do anything with it 🤔

It should be much simpler: if we don't have submissions, fetch the data.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,109 @@
...
+interface Props {
+  submissions: any;
+  dispatch: any;
+}
+
+class Submissions extends React.Component<Props, State> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      loading: true,
```
iampava, line 28: If we have submissions this will be false, so let's initialize it according to that.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,109 @@
...
+      this.setState({
+        loading: false,
+      });
+    }
+  }
+
+  getData = async () => {
+    const { dispatch } = this.props;
+    try {
+      await ExerciseService.getSubmissions(0, '')
```
iampava, line 54: You either use `async-await` or `Promises and .then`. With a few exceptions we usually don't combine the both, which you do here.

This leads me to believe you haven't quite understood `async-await` yet. :D Give it another read and then look over this code again.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,109 @@
...
+        <Head>
+          <title>HTML</title>
+          <link rel="icon" href="/favicon.ico" />
+        </Head>
+        <Header />
+        <main className={styles.page}>
+          <div className={`${style['page-submissions']}`}>
+            <ul className={style.cardWrapper}>
+              {submissions.submissions !== undefined
+                ? submissions.submissions.map((submission) => (loading
```
iampava, line 79: This doesn't work, I never see the skeletons even if I specifically delay the response.

There's a big flaw in this code, I'm gonna let you look for it first and if you don't find it I'll point you in the right direction :D

iampava (changes requested):
Apart from the specific feedback, there's also this empty file here... Not sure what's going on with it...?

image

frontend-ssr/pages/submissions.module.scss
andreeatoma: Yes..because I also did the search component. It should be in Search.module.scss. I thought it should be no problem if I let the code here for now, but I will remove it.

iampava: Then let's add the search component as part of this PR as well.

andreeatoma: oky

frontend-ssr/pages/searchExerciseSubmissions.tsx
```diff
@@ -0,0 +1,43 @@
...
+  onSearchSubmission = async (e) => {
+    const { dispatch } = this.props;
+    e.preventDefault();
+    let keyword = e.target.searchKeyword.value;
+    try {
+      const response = await ExerciseService.getSubmissions(0, keyword);
+      const jsonResp = await response.json();
+      return dispatch(searchSubmissions(keyword, jsonResp));
+    } catch (err) {
+      return SweetAlertService.toast({ type: 'error', text: err });
```
iampava, line 25: Just a small nitpick, but no need for the return here.

frontend-ssr/pages/searchExerciseSubmissions.tsx
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable @typescript-eslint/no-empty-interface */
```
iampava, line 1: This should be a generic component. We're gonna re-use it in multiple pages so it shouldn't what we're searching.

frontend-ssr/pages/submissions.module.scss
iampava: Sorry, I didn't realize it yesterday, but if it's a separate component, it should have separate styles in it's own `.module.scss` file.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Head from 'next/head';
+import Footer from '~/components/footer/Footer.component';
+import Header from '~/components/header/Header.component';
+
+import styles from '../styles/pages/default-page.module.scss';
```
iampava, line 7: Why are we importing those styles?

Also as a general rule, let's always add the styles import last, just like in the other files, separated by a newline from the other import. It's easier to see what styles are we getting.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      loading: false,
+    };
+  }
+
+  componentDidMount() {
+    let { submissions } = this.props;
+    const state = submissions;
```
iampava, line 35: Not sure why you create a separate variable? We can directly do

```
const { submissions } = this.props;

if(submissions.submissions === undefined) {
  this.getData()
}
```

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
...
+    if (state.submission === undefined) {
+      this.getData();
+    }
+  }
+
+  getData = async () => {
+    const { dispatch } = this.props;
```
iampava, line 42: More of a styling rule and personal preference, but let's add a newline after declarations. Makes things easier to read :D

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
...
+interface Props {
+  submissions: any;
+  dispatch: any;
+}
+
+class Submissions extends React.Component<Props, State> {
+  constructor(props) {
+    super(props);
+    this.state = {
+      loading: false,
```
iampava, line 29: I had a previous comment here but it seems nothing changed.

Loading should be based on whether or not we have submissions. In other words:

```
this.state = {
  loading: props.submissions.submissions === undefined
}
```

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const state = submissions;
+    if (state.submission === undefined) {
+      this.getData();
+    }
+  }
+
+  getData = async () => {
+    const { dispatch } = this.props;
+    try {
+      const response = await ExerciseService.getSubmissions(0, '');
```
iampava, line 44: I see we're always getting the very first page, ie: page=0. This doesn't seem right. We should get the "next page" if we already have some pages.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
...
+      const jsonResp = await response.json();
+      return dispatch(loadSubmissions(jsonResp));
+    } catch (err) {
+      return SweetAlertService.toast({ type: 'error', text: err });
```
iampava, line 48: Some here, a small nitpick but to keep things as minimal as possible, no need for `return`.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
...
+          <div className={`${style['page-submissions']}`}>
+            <SearchExerciseSubmissions />
+            <ul className={`${style['cards-wrapper']}`}>
+              {!loading && submissions.submissions !== undefined
+                ? submissions.submissions.map((submission: Submission) => (
+                  <li key={submission.id}>
+                    <ExSubmission submission={submission} />
+                  </li>
+                ))
+                : Array.from(Array(5), (i) => (
```
iampava, line 72: Much much better. This is exactly what I wanted to suggest, a fixed numbe of skeletons equal to the page size. Good job!

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <Head>
+          <title>HTML</title>
+          <link rel="icon" href="/favicon.ico" />
+        </Head>
+        <Header />
+        <main className={styles.page}>
+          <div className={`${style['page-submissions']}`}>
+            <SearchExerciseSubmissions />
+            <ul className={`${style['cards-wrapper']}`}>
+              {!loading && submissions.submissions !== undefined
```
iampava, line 66: I think only the second part of the `if` should be here. 

Think of the case where we're getting the second page. `loading` is true, we show the 5 skeletons at the end, but we still want to show the other submissions, right?

iampava (changes requested):

frontend-ssr/pages/search.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
iampava, line 1: Small nitpick: let's use PascalCase for the component name, ie `Search.tsx` and `Search.module.scss`.
Then also add another `index.ts` file in the same folder which exports that component so we can directly import it like this `import Search from 'components/Search` instead of `import Search from 'components/search/Search'`

iampava: For this pattern have a look at `Header`, `Footer`, etc. It's newly introduced, I saw it used by @catalinpopusoi  and I think it's a good idea which makes for cleaner imports.

frontend-ssr/pages/search.tsx
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  render() {
+    const { searchValue } = this.state;
+    const { onSearch } = this.props;
+
+    return (
+      <div>
+        <form
+          onSubmit={(e) => onSearch(e)}
```
iampava, line 38: Small nitpick but we can directly do `onSubmit={onSearch}`.

frontend-ssr/pages/submissions.module.scss
```diff
@@ -0,0 +1,14 @@
+@import "../styles/variables";
+.page-submissions {
+  width: $width;
```
iampava, line 3: Let's extract those 3 rules in a placeholder - what we're already doing with all the gradients. And then here and in the `Lesson.module.scss` we just inlclude it. This way we remove some duplicate styles.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Head from 'next/head';
+import Footer from '~/components/footer/Footer.component';
+import Header from '~/components/header/Header.component';
+
+import ExSubmission from '../components/exercise-submissions/ExerciseSubmission';
```
iampava, line 7: As a rule of thumb, always name the import exactly like the file name.  `ExerciseSubmissions`, `ExerciseSubmissionsSkeleton`, etc. Although this means a little more code, it helps avoid some confusion because everything is named the same way.

I don't have to ask myself: is `ExSubmission` same as `ExerciseSubmission` or a little different?

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,117 @@
...
+      this.getData();
+    }
+  }
+
+  getData = async () => {
+    const { submissions, dispatch } = this.props;
+
+    try {
+      const response = await ExerciseService.getSubmissions(submissions.page, '');
+      const jsonResp = await response.json();
```
iampava, line 47: This `.json` should be made in the service. Otherwise we'll need to call it every time.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,117 @@
...
+    this.state = {
+      // eslint-disable-next-line react/no-unused-state
+      loading: props.submissions.submissions === undefined,
+    };
+  }
+
+  componentDidMount() {
+    let { submissions } = this.props;
+
+    if (submissions.submission === undefined) {
```
iampava, line 37: Typo; it should be `props.submissions.submissions`. Right now if you navigate back-forward it will re-fetch another page of submissions although it shouldn't.

Also, this should have been caught by TypeScript. I see you didn't correctly declare prop types, both are any. It should be the correct types.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,117 @@
...
+  dispatch: any,
+}
+
+class Submissions extends React.Component<Props, State> {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      // eslint-disable-next-line react/no-unused-state
+      loading: props.submissions.submissions === undefined,
```
iampava, line 30: I don't see you using this variable anywhere. It should work like this:

While the submissions are fetched we show another 5 skeletons at the end. When the results came back or if an error happened, hide the skeletons.

iampava (changes requested):
A few improvements but the overall redux use is great! Good job!

frontend-ssr/components/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+
+import styles from './Search.module.scss';
+
+interface State {
+  searchValue: string;
+}
+interface Props {
+  query: string;
+  key: string;
```
iampava, line 10: This isn't needed, right?

frontend-ssr/components/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
iampava, line 1: Nitpick: the proper name should be capitalized - `components/Search` - just like `Header` and `Footer`.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,134 @@
...
+    } catch (err) {
+      SweetAlertService.toast({ type: 'error', text: err });
+      this.setState({
+        loading: false,
+      });
+    }
+  };
+
+  searchData = async (e: any) => {
+    e.preventDefault();
```
iampava, line 70: This prevent default should be in `Search` component. Otherwise we'll have to add `e.preventDefault` in every component that uses search. 

Also, the search should call this function directly with the query string as the parameter. So that here we don't have to do `e.target.query.value`.

frontend-ssr/components/search/Search.tsx
```diff
@@ -0,0 +1,58 @@
...
+          onSubmit={onSearch}
+          className={`${styles.form} d-flex align-items-center`}
+        >
+          <input
+            type="text"
+            name="query"
+            placeholder="search"
+            value={searchValue}
+            onChange={this.onChange}
+            required
```
iampava, line 47: This should not be required. Otherwise we're not going to be able to reset the search.

frontend-ssr/pages/submissions.tsx
iampava: I see you've marked this comment as resolved but the `.json` is still here 🤔

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,134 @@
...
+import Search from '../components/Search';
+import { RootState } from '../redux/root.reducer';
+
+import styles from './submissions.module.scss';
+
+interface State {
+  loading: boolean;
+}
+
+const mapState = (state: RootState) => ({
```
iampava, line 24: Move this one at the end as well, just like the connector.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+    const { dispatch, submissions } = this.props;
+    const keyword = e.target.query.value;
+
+    try {
+      const response = await ExerciseService.getSubmissions(
+        submissions.page,
+        keyword,
+      );
+      const jsonResp = await response.json();
```
iampava, line 80: `.json` here as well. Should be made in the service so we don't have duplicate code.

frontend-ssr/services/ExerciseSubmissions.service.ts
```diff
@@ -5,7 +5,8 @@ import HttpService from './Http.service';
 
 class ExerciseService {
   static getSubmissions(page = 0, query = '') : Promise<Submission[]> {
-    return HttpService.get(`${process.env.ENDPOINT}/exercises/submissions?page=${page}&query=${query}`) as unknown as Promise<Submission[]>;
+    return HttpService.get(`${process.env.ENDPOINT}/exercises/submissions?page=${page}&query=${query}`)
+      .then((response) => response) as unknown as Promise<Submission[]>;
```
iampava, line 9: Not sure what you wanted to do here but you added a `.then` which just returns the `response` which is basically useless... You should have called `..json` on the response and return that.

iampava (changes requested):

PR #38, opened 2020-11-05
Infinite Scroll and Realtime Search

andreeatoma:


frontend-ssr/components/search/Search.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import _ from 'lodash';
```
iampava, line 2: Let's use the `lodash-es` module and import just what we need:

`import { debounce } from 'lodash-es'`.

This way we only add to the final bundle the code we need, not all of lodash which is huge - https://bundlephobia.com/result?p=lodash@4.17.20

iampava (changes requested):
Something is a little wrong with the PR.

A lot of files appear inside `Files Changed` tab although they exist already inside `iampava:exercise-submissions`. Please have a sync first, maybe it will fix things.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,161 @@
...
+      dispatch(searchSubmissions(query, newSubmissions));
+    } catch (err) {
+      SweetAlertService.toast({ type: 'error', text: err });
+    } finally {
+      this.setState({ loading: false });
+    }
+  };
+
+  loadMore = (entries) => {
+    let { prevY } = this.state;
```
iampava, line 88: I'm a little confused about the code here. 

You should only check the `isIntersecting` property on the entry, without any additional math. If that's true, load some more elements.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,161 @@
...
+    if (submissions.submissions === undefined) {
+      this.getData();
+    }
+
+    const options = {
+      threshold: 0.7,
+    };
+
+    this.observer = new IntersectionObserver(this.loadMore, options);
+    this.observeNextPage();
```
iampava, line 51: I don't think we need a separate function for this. Instead I would create an: `initIntersectionObserver` function with the `options` object and the initialization code. It would make `componentDidMount` cleaner.

frontend-ssr/pages/submissions.tsx
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  componentWillUnmount() {
+    this.observer.disconnect();
```
iampava, line 55: Kudos for disconnecting when we leave.

frontend-ssr/pages/submissions.tsx
iampava: Also, let's not observe the button, it happens to be at the end but that's not always a guarantee.

Instead I would add an "invisible" element at the end which we'll observe for intersections.

iampava (changes requested):

frontend-ssr/pages/submissions.tsx
iampava: However, we should also disconnect when we reach the end: meaning we have no data left on the server.

For this we should add a new property in store, which tells us if we're at the end or not. This way, if we navigate back-forward to this page, it will know we're still at the end and **should not** initialize the Observer.

frontend-ssr/pages/submissions.tsx
iampava: Another point here is that we should only initialize the observer if we haven't reached the end. See another comment down  below regarding that.

frontend-ssr/redux/exercise-submissions/exercise-submissions.reducer.ts
```diff
@@ -53,6 +54,7 @@ export const submissionReducer = (state = initialState, action: { type: string;
           ? [...state.submissions, ...newSubmissions]
           : newSubmissions,
         page: state.page + 1,
+        end: true,
```
iampava, line 57: Why are we hardcoding `end: true` here? This means that after the first fetched page we're done.

But what if there are more pages?

frontend-ssr/pages/submissions.tsx
```diff
@@ -36,6 +42,27 @@ class Submissions extends React.Component<ConnectedProps<typeof connector>, Stat
     if (submissions.submissions === undefined) {
```
iampava, line 42: Now that we have the intersection observer logic, we can get rid of this code:

```
  let { submissions } = this.props;
    if (submissions.submissions === undefined) {
      this.getData();
    }
```

and just fetch based on the intersection, because  when you enter the page -  if there are NO elements it will be intersecting and start the fetch. Otherwise (if it's a return visit) - there will be elements and no fetch will happen.

For this to work you'll also need to revert 

```
this.state = { loading: false }
```

frontend-ssr/pages/submissions.tsx
```diff
@@ -36,6 +42,27 @@ class Submissions extends React.Component<ConnectedProps<typeof connector>, Stat
...
       this.getData();
     }
+
+    this.initIntersectionObserver = () => {
```
iampava, line 46: Why are we dynamically creating this function instead of adding it on the class body?

iampava: Also, this function should also start observing if we've not reached the end yet.

frontend-ssr/pages/submissions.tsx
```diff
@@ -67,6 +94,20 @@ class Submissions extends React.Component<ConnectedProps<typeof connector>, Stat
...
   };
 
+  loadMore = (entries) => {
+    const { submissions } = this.props;
+
+    entries.forEach((entry) => {
+      if (entry.isIntersecting) {
+        this.getData();
+
+        if (submissions.end) {
```
iampava, line 104: This disconnect code should not be in this function. Instead it should be extracted outside,probably in a `componentDidUpdate`.

iampava (changes requested):

frontend-ssr/pages/submissions.tsx
```diff
@@ -22,22 +22,49 @@ interface State {
...
 
     this.state = {
-      loading: props.submissions.submissions === undefined,
+      loading: false,
     };
   }
 
   componentDidMount() {
-    let { submissions } = this.props;
+    this.initIntersectionObserver();
```
iampava, line 38: You should only initialize it if we haven't reached the end.  Otherwise, whenever you go back-forward, it will try and fetch the next page.

frontend-ssr/pages/submissions.tsx
```diff
@@ -22,22 +22,49 @@ interface State {
...
+  }
+
+  componentDidUpdate(prevProps) {
+    const { submissions } = this.props;
 
-    if (submissions.submissions === undefined) {
-      this.getData();
+    if (submissions !== prevProps.submissions && submissions.end) {
```
iampava, line 44: The first part of the `if` is useless. Everytime `submissions` will be different from `prevProps.submissions` due to the nature of Redux which always returns new objects.

You should check whether we reached the end and the previous value was false

`prevProps.submissions.end !== submissions.end && submissions.end`

frontend-ssr/pages/submissions.tsx
```diff
@@ -22,22 +22,49 @@ interface State {
...
+  }
+
+  initIntersectionObserver = () => {
+    const options = {
+      threshold: 0.3,
+    };
+
+    this.observer = new IntersectionObserver(this.loadMore, options);
+
+    const elToObserve = this.hiddenRef.current;
```
iampava, line 60: We can delete all the code below and replace directly with `this.observer.observe(this.hiddenRef.current);` since we know the hiddenRef will always be there?

frontend-ssr/pages/submissions.tsx
iampava: Also, make sure the `disconnect` inside `componentWillUnmount` also check if this observer has been initialized.

iampava (changes requested):

frontend-ssr/components/search/Search.tsx
```diff
@@ -1,4 +1,6 @@
 import React from 'react';
```
iampava, line 1: Something is wrong with the search, it doesn't work. It seems to be making 2 requests, and the last one is without `query` param.

iampava: When you fix it test this scenario with a lower PAGE_SIZE so we have more than 1 page:

1. Go on page
2. Search some mambo-jumbo string that has no results  -> should see 0 results in page and the observer should be disconnected
3. Delete the search string -> first page of results should appear
4. Scroll to the bottom -> second page should be fetched

frontend-ssr/pages/submissions.tsx
andreeatoma: yep. we know :)

PR #40, opened 2020-11-25
[NotificationsComponent] functionality

andreeatoma:


frontend-ssr/components/Scroll/Scroll.tsx
```diff
@@ -0,0 +1,204 @@
+/* eslint-disable no-unused-expressions */
```
iampava, line 1: We should uncomment these rules and change the code in such a way that it doesn't conflict. Don't be afraid to modify the original code - that was written in a rush and without TypeScript.

frontend-ssr/components/Scroll/Scroll.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  _disposeEndReachEvents() {
+    this.bScroll.off('scroll', this._checkEndReach);
+    this.bScroll.off('touchEnd', this._checkEndReach);
+  }
+
+  _checkEndReach() {
+    const { onEndReach } = this.props;
+
+    let preloadDistance = this.props.preloadDistance || INFINITE_SCROLL_DISTANCE;
```
iampava, line 130: You forgot to destructure `preloadDistance`. ESLint complains here

frontend-ssr/components/Scroll/Scroll.tsx
```diff
@@ -0,0 +1,204 @@
...
+          !this.forceStop.end && this.bScroll.on('scroll', this._checkEndReach);
+        })
+        .catch(() => this.bScroll.refresh());
+    }
+  }
+
+  _checkStartReach() {
+    const { onStartReach } = this.props;
+
+    let preloadDistance = this.props.preloadDistance || INFINITE_SCROLL_DISTANCE;
```
iampava, line 148: Forgot destructuring. ESLint complains

frontend-ssr/components/Scroll/Scroll.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  scrollRef: React.RefObject<HTMLDivElement>;
+
+  bScroll: any;
+
+  constructor(props) {
+    super(props);
+    this.forceStop = {};
+    this.scrollRef = React.createRef();
+    this._checkStartReach = throttle(this._checkStartReach.bind(this), 50);
```
iampava, line 37: We can avoid `bind` here. When I wrote this code in DevDrive I didn't know that using 

`_checkStartReach = (...) => { ... }` will preserve the `this`. So using this method - called `class fields` if I'm not mistaken - makes things more readable.

frontend-ssr/components/Scroll/Scroll.tsx
```diff
@@ -0,0 +1,204 @@
...
+      maxY: this.bScroll.maxScrollY,
+    };
+  }
+
+  _disposeScrollEvents() {
+    this._disposeStartReachEvents();
+    this._disposeEndReachEvents();
+  }
+
+  _setupScrollEvents() {
```
iampava, line 94: Let's rename all functions starting with `_` by removing the `_`. Since in DevDrive I haven't got TypeScript - I used this notation to indicate it's supposed to be a private function. But here we can actually use the `private` descriptor.

frontend-ssr/redux/user/types.ts
```diff
@@ -0,0 +1,28 @@
+export interface UserState {
+  info: {
+    name: 'string',
+    username: 'string',
+    email: 'string',
+    avatar: 'string',
+  }
+  notifications: Notifications;
+  page: number,
```
iampava, line 9: `page` and `end` don't make sense to be on the `user` directly. They should be inside `notifications`, since they relate to that.

frontend-ssr/services/User.service.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  static async getNotifications(page = 0) {
+    await UserService.wait(LOAD_DURATION);
+    return MOCK_NOTIFICATIONS;
+  }
+
+  static async markAsRead(id) {
+    await UserService.wait(LOAD_DURATION);
+    return MOCK_NOTIFICATION_READ;
```
iampava, line 19: Not sure why this function and the below one are returning something. We don't expect/need any data when marking as read.

frontend-ssr/services/User.service.ts
```diff
@@ -0,0 +1,28 @@
+import MOCK_NOTIFICATIONS from '~/pages/api/notifications/mock';
+import MOCK_NOTIFICATION_READ from '../pages/api/notifications/mockAsRead';
+import MOCK_NOTIFICATIONS_ALL_READ from '../pages/api/notifications/mockAllAsRead';
+
+const LOAD_DURATION = 1000;
+
+class UserService {
+  static wait(mili) {
+    return new Promise((resolve) => setTimeout(resolve, mili));
```
iampava, line 9: Nice use of `wait` to simulate the delay.

frontend-ssr/styles/_animations.scss
```diff
@@ -21,3 +21,26 @@
     opacity: 0.5;
   }
 }
+
+@keyframes dummyNotificationAnimation {
```
iampava, line 25: We should use the same `skeletonOpacity` animation for the Notification Skeleton. Let's keep it consistent across the app - less code to write.

frontend-ssr/redux/user/user.reducer.ts
```diff
@@ -0,0 +1,84 @@
...
+};
+
+export const userReducer = (state = initialState, action: { type: string; payload: any; })
+  : UserState => {
+  switch (action.type) {
+    case USER_NOTIFICATIONS.ADD: {
+      let { index, notification } = action.payload;
+
+      if (!Array.isArray(notification)) {
+        notification = [notification];
```
iampava, line 18: If this is an `array` it should be in plural - `notifications`

frontend-ssr/redux/user/user.reducer.ts
```diff
@@ -0,0 +1,84 @@
...
+      };
+    }
+    case USER_NOTIFICATIONS.MARK_AS_READ: {
+      let { id } = action.payload;
+
+      return {
+        ...state,
+        notifications: {
+          ...state.notifications,
+          list: [{ ...state.notifications.list.find((n) => n._id === id), isUnread: false }],
```
iampava, line 53: This cannot be correct. `list` is an Array of elements, but if you're using `find` you're just returning one element.

So basically, you're erasing the entire list, and keeping just one.

frontend-ssr/redux/user/user.reducer.ts
```diff
@@ -0,0 +1,84 @@
...
+      };
+    }
+    case USER_NOTIFICATIONS.MARK_AS_UNREAD: {
+      let { id } = action.payload;
+
+      return {
+        ...state,
+        notifications: {
+          ...state.notifications,
+          list: [{ ...state.notifications.list.find((n) => n._id === id), isUnread: true }],
```
iampava, line 65: Same here. You should find a way to modify just one notification, while keeping the rest.

frontend-ssr/pages/api/notifications/mockAsRead.ts
```diff
@@ -0,0 +1,56 @@
+import { uuid } from '~/services/Utils';
```
iampava, line 1: We shouldn't need this file. We don't return anything when marking a notification as read.

frontend-ssr/pages/api/notifications/mockAllAsRead.ts
```diff
@@ -0,0 +1,57 @@
+import { uuid } from '~/services/Utils';
```
iampava, line 1: We shouldn't need this file. We don't return anything when marking all notifications as read.

frontend-ssr/pages/api/notifications/mock.ts
```diff
@@ -0,0 +1,57 @@
+import { uuid } from '~/services/Utils';
+import { Notification } from '../../../redux/user/types';
+
+const MOCK_NOTIFICATIONS: Notification[] = [
```
iampava, line 4: Good job with the mocked info, but let's not put it in the `api` folder. Move it in `UserService.ts` since it's the only place we use it. So that we only modify thing in one file when needed.

frontend-ssr/components/notifications-tooltip/loading-notification/LoadingNotification.component.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
iampava, line 1: Let's rename to `NotificationSkeleton.tsx` - to keep the same naming convention as in `ExerciseSubmissionSkeleton`.

frontend-ssr/components/notifications-tooltip/loading-notification/LoadingNotification.module.scss
```diff
@@ -0,0 +1,29 @@
+@import "../../../styles/variables";
```
iampava, line 1: We might not need this file after we extract the single Notification in a separate component - just like we don't have a separate styling for the `ExerciseSubmissionSkeleton.tsx`.

frontend-ssr/components/notifications-tooltip/NotificationsTooltip.component.tsx
```diff
@@ -0,0 +1,221 @@
+import React from 'react';
```
iampava, line 1: This should be named `NotificationTooltip` - just like the folder to keep the same convention as the other components.

frontend-ssr/components/notifications-tooltip/NotificationsTooltip.component.tsx
```diff
@@ -0,0 +1,221 @@
...
+            {notifications.map(({
+              _id, username, name, avatar, message, timestamp, url, isUnread,
+            }) => (
+              <li
+                key={_id}
+                className={isUnread ? `${styles['is--unread']}` : ''}
+                onClick={() => this.markAsRead(_id)}
+                role="presentation"
+              >
+                <Link href="/">
```
iampava, line 177: Let's extract the individual notification component into a separate file.

frontend-ssr/components/notifications-tooltip/NotificationsTooltip.component.tsx
```diff
@@ -0,0 +1,221 @@
...
+  markAsRead = async (id) => {
+    const { notifications } = this.props.user;
+    const { dispatch } = this.props;
+
+    if (notifications.list.find((n) => n._id === id).isUnread) {
+      try {
+        await UserService.markAsRead(id);
+        dispatch(markNotificationAsRead(id));
+      } catch (err) {
+        dispatch(markNotificationAsUnread(id));
```
iampava, line 119: Great great job for marking the notification back to `Unread` if the request fails 🎉🎉🎉

iampava (changes requested):

frontend-ssr/components/Scroll/Scroll.tsx
andreeatoma: Yes, I have recently read about this method. :D Class fields it is.

frontend-ssr/redux/user/types.ts
andreeatoma: yeah, as in the example with submissions.

frontend-ssr/services/User.service.ts
andreeatoma: Yes, you're right, if you remember I asked you a question about why we need 3 API's. In fact, I meant the 3 mocking files.

frontend-ssr/components/notification-tooltip/Notification.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import Link from 'next/link';
+
+import { timeAgo } from '../../services/Utils';
+import styles from './Notification.module.scss';
+
+interface Props {
+  id: string,
```
iampava, line 8: `id` is useless here. We're not using it inside this component so let's delete it.

frontend-ssr/components/notification-tooltip/NotificationTooltip.component.tsx
```diff
@@ -0,0 +1,234 @@
...
+                  url,
+                  isUnread,
+                }) => (
+                  <li
+                    key={_id}
+                    className={isUnread ? `${styles['is--unread']}` : ''}
+                    onClick={() => this.markAsRead(_id)}
+                    role="presentation"
+                  >
+                    <Notification
```
iampava, line 196: Let's simplify here. Instead of passing each prop one by one we can take advantage of destructuring:

```
{notifications.map(notificationProps => (
  <li>
     <Notification {...notificationProps} />
  </li>
))}
```

frontend-ssr/components/Scroll/Scroll.tsx
iampava: You still left one of them 😅

frontend-ssr/services/User.service.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  static async getNotifications(page = 0) {
+    await UserService.wait(LOAD_DURATION);
+    return MOCK_NOTIFICATIONS;
+  }
+
+  static async markAsRead(id) {
+    console.log(`The notification with id ${id} was read`);
```
iampava, line 70: Having a `console.log` is not a good mocking technique. We want something asyncronous, to emulate a delay.

We should have:

`await UserService.wait(LOAD_DURATION`)`

Same mention in the below function.

frontend-ssr/components/notifications-tooltip/loading-notification/LoadingNotification.component.tsx
iampava: I see you renamed the file, kudos for that. But I kinda expected you to also re'arrange the folders in a similar fashion with ExerciseSubmission. What I mean is:

```
📁 notification-tooltip
      NotificationTooltip.tsx
      📁notification
           Notification.tsx
           NotificationSkeleton.tsx
           Notification.module.scss
```

I know this may not seem like a big deal, and if it were just us coding it probably wouldn't be. But since I expect more contributions, it's important to have the same structure everywhere.

frontend-ssr/components/notification-tooltip/Notification.module.scss
```diff
@@ -0,0 +1,135 @@
...
+      margin-right: 0.4em;
+    }
+  }
+
+  .hidden {
+    visibility: hidden;
+  }
+}
+
+.dummy-notification {
```
iampava, line 105: Let's rename this to `skeleton` as in `ExerrciseModule.scss`. Also make sure there are as little duplicate styles as possible between the normal notification and the skeleton one.

iampava (changes requested):

iampava:
I cherry picked this commit because I wanted to do some small changes, and merged it separately into master.

PR #42, opened 2020-12-08
Images Lesson small changes

andreeatoma:


frontend-ssr/components/SkeletonImage/SkeletonImage.module.scss
```diff
@@ -0,0 +1,23 @@
...
+  &:after {
+    content: '';
+    width: 100%;
+    height: 0px;
+    display: block;
+    position: absolute;
+    box-shadow: 0 0 1px 3px lighten(color(grey), 20%);
+    top: 50%;
+    left: 0;
+    animation: skeletonOpacity 2s ease-in-out infinite;
```
iampava, line 21: This animation is not working.

frontend-ssr/components/SkeletonImage/SkeletonImage.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import styles from './SkeletonImage.module.scss';
+
+interface Props {
+  width: string,
+  height: string,
+}
+
+const SkeletonImage = ({ width, height } : Props) => (
+  <li className={styles.skeleton}>
```
iampava, line 10: We should not render a `<li>` here as this forces us to always have an `ul` or `ol` as a parent.  Just render the `div` and depending on the use case we will wrap it into a `<li>` from the outside.

frontend-ssr/components/SkeletonImage/SkeletonImage.module.scss
```diff
@@ -0,0 +1,23 @@
+@import "../../styles/variables";
```
iampava, line 1: You forgot an `index.ts` file inside the `SkeletonImage` folder - take a look at the `Header` component for example and how it's imported.

frontend-ssr/components/lessons/LessonComments/LessonComments.module.scss
```diff
@@ -0,0 +1,5 @@
+@import '../../../styles/variables';
```
iampava, line 1: I think we can get rid of this, right? 

Or, if you to add a border for testing/development purposes - add it inline so we don't need to add another file, just for that. And we'll delete it on the final integration.

frontend-ssr/components/lessons/LessonComments/LessonComments.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import { connect, ConnectedProps } from 'react-redux';
+import { RootState } from '~/redux/root.reducer';
+
+import {
+  loadComments,
+} from '../../../redux/lesson-comments/lesson-comments.actions';
```
iampava, line 7: You can import them using `~` alias which should result in a smaller path. Same for the `LessonCommentsService`.

It's not a big deal but if we ever move this component, we will not have to re-write all the paths.

frontend-ssr/components/lessons/LessonComments/comment/CommentBlock.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <a href="#" className="d-flex no-underline">
+          <img
+            width="32"
+            height="32"
+            src={avatar}
+            alt={`${username} avatar`}
+          />
+          <span className={styles.user}>{name || username}</span>
+        </a>
+        <span> a lasat urmatorul comentariu: </span>
```
iampava, line 33: I looked over this UI and let's

1. Let's make the image a little bigger
2. The image should be in a link that leads to the user profile
3. Instead of the "a lasat urmatorul comentariu" let's put the username - also a link to the user profile. Here maybe you can combine them in the same link (the image and the username text)

frontend-ssr/components/lessons/LessonComments/newComment/NewComment.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+   render() {
+     const { newComments } = this.state;
+
+     return (
+       <div className={styles['new-comment']}>
+         {newComments.map((newComment) => <CommentBlock key={newComment._id} {...newComment} />)}
+         <form onSubmit={this.handleSubmit}>
+           <div>
+             <input
```
iampava, line 64: Regarding the inputs let's just have a textarea with the actual comment for now as we don't need the rest of the inputs.

For this story let's mock the UI, avatar and so on. For the final integration things will look like this:

1. User wants to comment
2. If it's already logged in - comment gets added
3. Otherwise, show login popup. User logs in/registers -> then the comment is added.

frontend-ssr/package.json
```diff
@@ -29,6 +29,7 @@
     "@types/mongoose": "^5.7.36",
     "@types/react-router": "^5.1.8",
     "@types/uuid": "^8.3.0",
+    "better-scroll": "^2.0.5",
```
iampava, line 32: I don't think we need those anymore, I removed the scroll for now.

frontend-ssr/components/lessons/LessonComments/newComment/NewComment.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+     form.reset();
+   };
+
+   render() {
+     const { newComments } = this.state;
+
+     return (
+       <div className={styles['new-comment']}>
+         {newComments.map((newComment) => <CommentBlock key={newComment._id} {...newComment} />)}
```
iampava, line 61: Why  are we mapping here?

frontend-ssr/components/lessons/LessonComments/newComment/NewComment.tsx
```diff
@@ -0,0 +1,109 @@
...
+       avatar,
+       username,
+       name,
+       comment,
+       timestamp: Date.now(),
+     };
+
+     dispatch(addComment(newComment));
+
+     this.setState({
```
iampava, line 49: This doesn't look right. Why are we adding the comment to the local state?

frontend-ssr/redux/lesson-comments/lesson-comments.reducer.ts
```diff
@@ -0,0 +1,49 @@
+import { CommentState } from './types';
+import { LESSON_COMMENTS } from './lesson-comments.actions';
+
+const initialState: CommentState = {
+  comments: undefined,
```
iampava, line 5: No pagination, right? Just wanted to make sure.

It's ok, it's not as mandatory as for notifications/submissions since I don't expect a lot of comments in the beginning.

frontend-ssr/services/LessonComments.service.ts
```diff
@@ -0,0 +1,60 @@
...
+];
+
+const LOAD_DURATION = 1000;
+
+class LessonCommentsService {
+  static wait(mili) {
+    return new Promise((resolve) => setTimeout(resolve, mili));
+  }
+
+  static async getComments(id) {
```
iampava, line 54: You should have another method here for adding a comment. Then in the component we "wait" for it to finish before dispatching the new comment.

iampava (changes requested):

frontend-ssr/redux/lesson-comments/types.ts
```diff
@@ -0,0 +1,12 @@
+export interface CommentState {
```
iampava, line 1: Just realized this comments state is not ok. With the current state we only save the comments of one lesson at a time. 

Ideally we would persist the comments, so that if you jump between lessons they are still there.  So the comments need to be related to a specific lesson - through a `lessonId`.

iampava (changes requested):
I didn't see it the first time I reviewed this code, but the comments need to be tied to a specific lesson (see my remark regarding the store)

Other than this, only minor issues 👏

PR #60, opened 2021-01-11
[BUGFIX][ImagesLesson] bugfixes 

andreeatoma:
…End-ro/frontend.ro/issues/59#

### DIFF
--- a/frontend-ssr/pages/demo/html/fixed-size-image.tsx
+++ b/frontend-ssr/pages/demo/html/fixed-size-image.tsx
@@ -25,12 +25,7 @@ export default function FixedSizeImage() {
         <Header demoPage />
         <Demo title={title}>
           <DemoPreview>
-            <header style={{
-              display: 'flex',
-              alignItems: 'center',
-              justifyContent: 'space-between',
-            }}
-            >
+            <header className="d-flex align-items-center justify-content-between flex-row flex-wrap">
               <img
                 height="70"
                 width="212"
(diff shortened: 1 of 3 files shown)


frontend-ssr/pages/demo/html/fixed-size-image.tsx
```diff
@@ -29,6 +29,7 @@ export default function FixedSizeImage() {
               display: 'flex',
```
iampava, line 29: I think this can be slightly refactored to use classes:`d-flex`, `align-items-center` etc. 

I know I wrote the code that way,  but the takeaway here is if you see something that can be improved, don't hesitate to improve it.

iampava (changes requested):
There's no need to include the URL to the issue in the commit name. We linked this PR with the issue directly in GitHub.

Also, for small fixes like this one, it's easier for me to review if you include images. A short print-screen: before and after. This way I don't have to fetch your branch, and run the code. I can directly approve it :D

frontend-ssr/pages/demo/html/fixed-size-image.tsx
andreeatoma: Ok, got it.

andreeatoma:
> There's no need to include the URL to the issue in the commit name. We linked this PR with the issue directly in GitHub.
> 
> Also, for small fixes like this one, it's easier for me to review if you include images. A short print-screen: before and after. This way I don't have to fetch your branch, and run the code. I can directly approve it :D

before

104126663-8189e680-5366-11eb-859f-318bc1240c5c
104126659-7c2c9c00-5366-11eb-811f-143cbfa961ec

after :D

image
image



> There's no need to include the URL to the issue in the commit name. We linked this PR with the issue directly in GitHub.
> 
> Also, for small fixes like this one, it's easier for me to review if you include images. A short print-screen: before and after. This way I don't have to fetch your branch, and run the code. I can directly approve it :D

Issue #74, opened 2021-01-22
Add logrocket for session capturing


nmaties:
Add logrocket for session capturing and other logs 
https://logrocket.com/

iampava:
As part of this issue please add a small privacy page. You can take inspiration from here: https://devdrive.io/privacy - this is another app of mine. 

What I wanna say is let's keep the privacy info to the point and  "human readable". Not like all those other Terms of Service :))

PR #77, opened 2021-01-26
[Analytics] Added logrocket

nmaties:
Closes: https://github.com/FrontEnd-ro/frontend.ro/issues/74

### DIFF
--- a/client/components/IdentifyLogRocket.tsx
+++ b/client/components/IdentifyLogRocket.tsx
@@ -1,0 +1,36 @@
+import { connect, ConnectedProps } from 'react-redux';
+import { RootState } from '~/redux/root.reducer';
+import LogRocketService from '~/services/utils/LogRocket.service';
+
+/**
+ * This component "listens" to the User Store and
+ * identifies the logged in user in LogRocket.
+ *
+ * When that user logs out, it clear all the data.
+ */
+function IdentifyLogrocket({ userInfo }: ConnectedProps<typeof connector>) {
+  if (typeof window === 'undefined') {
+    /** Client-side only */
+    return null;
+  }
+
+  if (!userInfo && LogRocketService.getIdentityData()) {
+    LogRocketService.anonymize();
+  }
+
+  if (userInfo && !LogRocketService.getIdentityData()) {
+    LogRocketService.identify(userInfo);
+  }
+
+  return null;
+}
+
+function mapStateToProps(state: RootState) {
+  return {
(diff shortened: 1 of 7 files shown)


frontend-ssr/pages/_app.tsx
```diff
@@ -12,6 +13,10 @@ export default function MyApp({ Component, pageProps }: any) {
     },
   });
 
+  if (typeof window !== 'undefined' && process.env.NODE_ENV === 'production') {
```
iampava, line 16: Can we use the `useEffect` hook here to simulate `componentDidMount`? This way we don't need to check for `window !== 'undefined'`

iampava (changes requested):

iampava:
@nmaties  can you also rename the commit acording to CONTIBUTING.md? Should be something like

`[Analytics] Added logrocket`

frontend-ssr/pages/_app.tsx
iampava: Also, I just realized that we should lazy load this and dynamically import it in the same `useEffect` hook. This way we can save 32KB - https://bundlephobia.com/result?p=logrocket@1.0.14 - from the initial bundle.

iampava: Don't forget about the "privacy" page as well. If logrocket is using cookies we should add it.

frontend-ssr/package.json
```diff
@@ -40,6 +40,7 @@
     "highlight.js": "^10.4.1",
```
iampava, line 40: You should still have the `yarn.lock` updated - meaning you should ru `yarn` inside the project and commit those changes as well.

frontend-ssr/pages/_app.tsx
```diff
@@ -12,6 +13,16 @@ export default function MyApp({ Component, pageProps }: any) {
     },
   });
 
+  useEffect(() => {
+    if (process.env.NODE_ENV === 'production') {
+      const runLogRocket = async () => {
```
iampava, line 18: I know TS doesn't allow async in `useEffect` but I think we should use Promises in this case. This might be confusing to someone who doesn't know  about this TS thing. In other words, at first view I might think: why is he creating a separate function and calling it.

Also, remember to handle the catch. A `console.error` is ok for now, we'll push those to the server at some point :D

iampava (changes requested):

PR #78, opened 2021-01-26
[Deployments] Automatic deployments twice per day via workflows in github

nmaties:
Created workflow for automatic deployments twice per day.


iampava:
Question: does this deploy twice per day no matter what or only if there are new commits in master?

Also, should we probably merge this into https://github.com/FrontEnd-ro/frontend.ro/tree/express-server branch ?

nmaties:
> Question: does this deploy twice per day no matter what or only if there are new commits in master?
> 
> Also, should we probably merge this into https://github.com/FrontEnd-ro/frontend.ro/tree/express-server branch ?

It's gonna deploy no matter what with the latest version of master, we can do it in such way so that it deploys only when we push to master.
It does not matter if we merge or not the express-server branch.

iampava:
> 
> 
> > Question: does this deploy twice per day no matter what or only if there are new commits in master?
> > Also, should we probably merge this into https://github.com/FrontEnd-ro/frontend.ro/tree/express-server branch ?
> 
> It's gonna deploy no matter what with the latest version of master, we can do it in such way so that it deploys only when we push to master.
> It does not matter if we merge or not the express-server branch.

What do you think is the best way? I feel like we should deploy on master commits, and not "waste resources" by deploying if there's no need, haha!

Regarding the second question. I'm asking because we don't have heroku yet on this branch and I suspect it will fail the deploy.

nmaties:
> > > Question: does this deploy twice per day no matter what or only if there are new commits in master?
> > > Also, should we probably merge this into https://github.com/FrontEnd-ro/frontend.ro/tree/express-server branch ?
> > 
> > 
> > It's gonna deploy no matter what with the latest version of master, we can do it in such way so that it deploys only when we push to master.
> > It does not matter if we merge or not the express-server branch.
> 
> What do you think is the best way? I feel like we should deploy on master commits, and not "waste resources" by deploying if there's no need, haha!
> 
> Regarding the second question. I'm asking because we don't have heroku yet on this branch and I suspect it will fail the deploy.

Got it, I can make it in such way that it will redeploy whenever we merge anything to master.
Sure, let's wait until we merge express-server branch.

iampava (approved):

PR #91, opened 2021-01-28
[BUGFIX][Preview Lesson Links and Buttons] added  between Texts and L…

andreeatoma:
Before changes

image

And this is the order after. Basically I added lesson "Links and Buttons" between Texts Lesson and Lists Lesson.

image


andreeatoma:
done, the lesson is live now.

Issue #63, opened 2021-01-14
New component for showcasing code demos with live preview


iampava:
Right now the `<Highlight>` component is used independently and underneath it we usually add pictures.

I think it would be nice if we would combine both, let them edit the example and show a live-preview.

iampava:
Closing for now. We have CodePen preview component.

PR #102, opened 2021-02-04
[LivePreviewDemoComponent] added 

andreeatoma:
Fixes https://github.com/FrontEnd-ro/frontend.ro/issues/63

### DIFF
--- a/client/components/Editor/BasicEditor/BasicEditor.tsx
+++ b/client/components/Editor/BasicEditor/BasicEditor.tsx
@@ -18,8 +18,9 @@ class BasicMonacoEditor extends MonacoBase {
     };
   }
 
-  componentDidMount() {
-    this.initEditor();
+  // eslint-disable-next-line react/static-property-placement
+  static defaultProps = {
+    showFileSwitcher: true,
   }
 
   initEditor = () => {
@@ -46,13 +47,19 @@ class BasicMonacoEditor extends MonacoBase {
 
   onModelChange = () => {
     const { folderStructure } = this.state;
+    const { onChange } = this.props;
+
     folderStructure.setContent(this.state.selectedFileKey, this.editor.getValue());
+
+    if (onChange) {
+      onChange(folderStructure);
+    }
 
     this.setState({ folderStructure });
   }
 
   render() {
-    const { readOnly } = this.props;
+    const { readOnly, showFileSwitcher } = this.props;
     const {
       folderStructure, fileSwitcherWidth, selectedFileKey, isDropable,
     } = this.state;
@@ -69,6 +76,8 @@ class BasicMonacoEditor extends MonacoBase {
       >
         {folderStructure.files.length || folderStructure.folders.length ? (
           <>
+            { showFileSwitcher
+            && (
             <FileSwitcher
               maxHeight={500}
               folderStructure={folderStructure}
@@ -84,6 +93,7 @@ class BasicMonacoEditor extends MonacoBase {
               onDownload={this.onDownload}
               onResize={this.onResize}
             />
+            )}
             <div
               className={`
                 ${styles.editor}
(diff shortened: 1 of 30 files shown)


client/components/LivePreviewDemo/IframeDemo/IframeDemo.tsx
```diff
@@ -0,0 +1,23 @@
...
+    let doc = iframe.contentDocument;
+    doc.open();
+    doc.write(sourceCode);
+    doc.close();
+    iframe.style.width = '100%';
+    iframe.contentWindow.document.body.style.color = 'silver';
+  };
+
+  return (
+    <iframe frameBorder="0" ref={writeLanguage} title="iframe" />
```
iampava, line 21: It's a little bit weird the way you're using refs here to modify the `iframe`. It works but seems like a hack.

A simpler solution would be to create the `ref` the usual way using `useRef`,  assign it to the iframe and then use the `useEffect` hook to trigger updates when the `sourceCode` variable changes.

client/components/Highlight/Highlight.module.scss
```diff
@@ -2,70 +2,8 @@
 
 .highlight {
   width: 100%;
-  margin-left: auto;
```
iampava, line 5: In the end we want to have both the `Highlight` component and the `IFrame` components. I know you deleted this code just to quickly test the ifame scenario :)

iampava (changes requested):

iampava:
PS: now that you're in the organization and have **write** access, you can make branches directly in this repo and make the PR this way.

client/components/Editor/BasicEditor/BasicEditor.tsx
```diff
@@ -18,9 +18,9 @@ class BasicMonacoEditor extends MonacoBase {
     };
   }
 
-  componentDidMount() {
-    this.initEditor();
-  }
+  // componentDidMount() {
```
iampava, line 21: Delete rather than comment it.

client/components/Editor/Monaco.base.tsx
```diff
@@ -10,7 +10,13 @@ import FolderStructure from '~/services/utils/FolderStructure';
 import SweetAlertService from '~/services/sweet-alert/SweetAlert.service';
 import { MONACO } from '~/services/Constants';
 
-class MonacoBase extends React.Component<any, any> {
+interface Props extends MonacoBase {
+  onChange: (folderStructure?: FolderStructure) => void,
+  [id: string]: any,
```
iampava, line 15: Move this one last. Let's start with the props we actually know are there.

client/components/LivePreviewDemo/IframeDemo/IframeDemo.tsx
```diff
@@ -0,0 +1,32 @@
...
+    doc.body.style.fontSize = '1.3em';
+
+    doc.body.innerHTML = sourceCode;
+  }, [sourceCode]);
+  return (
+    <iframe
+      frameBorder="0"
+      ref={iframeRef}
+      title="iframe"
+      style={{
```
iampava, line 21: This should be in a CSS module file.

iampava (changes requested):
Overall very good. 

I looked over the component and have some UI/UX feedback: 

1. You broke the component inside `exercitii/creeaza` It should have the FileSwitcher.
image

As a good practice, when modifying a component like this which is critical, check all pages which use that component and compare to the production environment. Should look and behave the same.


2. The suggestions are not showing in the LivePreview component, but they are in `exercitii/creeaza`

image

image

3. Move the file name in the other side please, for the LivePreview component

image

4. Add a prop to pass a description to this demo which we can add at the bottom:

image

iampava:
Closing this for now. 

We've integrated with CodePen for now, will get back to this PR when we decide to implement our own (custom design) live editor.

PR #104, opened 2021-02-07
Lesson/box model

catalinpopusoi:


client/components/demo/index.ts
```diff
@@ -1,8 +1,10 @@
 import Demo from './Demo/Demo';
 import DemoPreview from './DemoPreview/DemoPreview';
+import DevToolsClone from './DevToolsClone/DevToolsClone';
```
iampava, line 3: Let's put this in a different folder. The `demo` folder is only related to components we'll feature on the `/demo` pages. Like [https://frontend.ro/demo/html/atributul-start-pentru-liste-ordonate](https://frontend.ro/demo/html/atributul-start-pentru-liste-ordonate)

iampava (changes requested):
I didn't review the code yet, just how the component works. I have some UX feedback regarding it plus some small bugs.

1. Clicking the labels should focus the input. Maybe it would be a good idea to extend this to the entire rectangle for that specific part and change cursor to pointer + change bg color on hover
2. Inputs are buggy - it seems I cannot write a value bigger than 15. If that's intended le'ts increase the number. And when we pass that value let's show some message indicating this
3. Try writing letters in the input - it shows NaN
4. The inputs are "type=text". Shouldn't we have them "type=number"?
5. Content should specify width & height like in DevTools 

image

6. It should be centered when in vertical mode

image

7. The selected type is confusing. Right now the selected one is faded which made me think it's not the selected one. Let's use the checkbox we alrready have for this:

image

image

8. The rectangles should have the same size horizontally and vertically. In the below picture, those 2 lines should be equal:

image

iampava (approved):

client/components/demo/index.ts
MarianGeorgeMorosac: +1

iampava:
Looks great! Thanks!

iampava (approved):

PR #119, opened 2021-02-26
[EventsPage] created

andreeatoma:


andreeatoma:
included this one in [Events] & [EventDetails] pages PR, so we don't need it anymore.

PR #120, opened 2021-03-04
[Events] & [EventDetails] pages

andreeatoma:


iampava:
Added some more commits and merged separately as part of: https://github.com/FrontEnd-ro/frontend.ro/pull/123

PR #129, opened 2021-03-15
[RegisterEventCard] verify if the user is logged in and pre'fill the name & email

andreeatoma:
…name & email, but allow him  to change it.

### DIFF
--- a/client/components/event/RegisterEventCard/RegisterEventCard.tsx
+++ b/client/components/event/RegisterEventCard/RegisterEventCard.tsx
@@ -1,7 +1,9 @@
 import React, { useEffect, useState } from 'react';
+import { connect, ConnectedProps } from 'react-redux';
 import Link from 'next/link';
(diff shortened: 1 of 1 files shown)


client/components/event/RegisterEventCard/RegisterEventCard.tsx
```diff
@@ -153,7 +172,8 @@ function EventCard({
                 <span className="label text-bold mb-2">
                   Nume si prenume
                 </span>
-                <input type="text" name="name" required />
+                {userInfo && <input type="text" value={username} name="name" required onChange={onUsernameChange} />}
```
iampava, line 175: Let's not add local state for username & email. If we use `defaultValue` prop on inputs it should be enough. And we can get rid of setStates and onUsernameChange/onEmailChange.

iampava (changes requested):

iampava (approved):

Issue #130, opened 2021-03-19
Add statistics regarding events, participants, hours of training, slides, etc


iampava:
We want to create and use a StatsComponent with info regarding:
* events
* participants
* hours of training
* number of slides
* etc

One of the most common patterns for this is:

![image](https://user-images.githubusercontent.com/11327761/111755486-d172b580-88a1-11eb-90c4-9fa1f6c69524.png)

However, let's try and be creative first, see what other layouts/ideas we come up with and only if nothing comes to mind default to the most common pattern.

For inspiration we can search common patterns on Dribble: https://dribbble.com/search/participants. I also have an idea to make a grid and each "cell" is a different stat:

![image](https://user-images.githubusercontent.com/11327761/111756178-9755e380-88a2-11eb-9e0a-80d06e530d26.png)

Maybe we can try multiple things and see which one is best suited.

So, as a recap:
* try and be original. If we can't create something we like let's use the default pattern
* try and combine the Schmoes -> https://joeschmoe.io/ in this component. Not mandatory, but let's give it a try.
* add "funky" data besides the necessary one. Some things that come to mind: 
  *  number of times we write FrontEnd in our slides
  *  combined nr of pixels from all participant photos
  *  times we went over the planned duration
  * number of letters in our combined names
  * any other funky stat you can think of. It would be nice to add some personality 😆😆😆

iampava:
Done. Let's close!

PR #131, opened 2021-03-22
[StatsComponent] added animated counter stats

andreeatoma:
Implements https://github.com/FrontEnd-ro/frontend.ro/issues/130

### DIFF
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -1,16 +1,31 @@
+import {
+  faBook,
+  faCalendarCheck,
+  faUserAstronaut,
+  faUserNinja,
+} from '@fortawesome/free-solid-svg-icons';
+import { faGithub } from '@fortawesome/free-brands-svg-icons';
 import SEOTags from '~/components/SEOTags';
 import {
   LandingHero,
   LandingHowItWorks,
   LandingSubscribe,
   LandingThanks,
-  LandingAdBanner,
 } from '~/components/Landing';
 import Footer from '~/components/Footer';
+import PlatformStats from '~/components/PlatformStats';
 import { withClientMonitoring } from '~/services/Hooks';
 
 export default function Home() {
   withClientMonitoring();
+
+  const stats = [
+    { icon: faUserNinja, counter: 9, description: 'contribuitori cu inimă mare' },
+    { icon: faGithub, counter: 5, description: 'repo-uri open-source' },
+    { icon: faCalendarCheck, counter: 2, description: 'workshop-uri practice' },
+    { icon: faUserAstronaut, counter: 11, description: 'participanți la evenimente' },
+    { icon: faBook, counter: 5, description: 'lecții' },
+  ];
 
   return (
     <>
@@ -20,14 +35,13 @@ export default function Home() {
         description="Vrei să înveți FrontEnd? Aici ai parte de tutoriale gratuite și o comunitate de developeri care te vor ajuta să devii mai bun."
       />
       <>
-        {/* <LandingAdBanner /> */}
         <LandingHero />
         <LandingHowItWorks />
         <LandingThanks />
         <LandingSubscribe />
+        <PlatformStats stats={stats} />
         <Footer />
       </>
-
     </>
   );
 }
--- a/client/components/PlatformStats/CounterCard/CounterCard.module.scss
+++ b/client/components/PlatformStats/CounterCard/CounterCard.module.scss
@@ -1,0 +1,41 @@
+@import '../../../styles/variables';
+
+.counter-card {
+  width: 11em;
+
+  .icon-box {
+    width: 5em;
+    height: 5em;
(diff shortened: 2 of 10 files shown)


client/components/Landing/LandingStats/CounterCard/CounterCard.module.scss
```diff
@@ -0,0 +1,33 @@
...
+.counter-card {
+  width: 11em;
+  .icon-box {
+    width: 5em;
+    height: 5em;
+    margin: 1em auto;
+    border: 2px solid color(white);
+    transform: rotate(45deg);
+    transition: transform 2s ease-out 100ms;
+    &:hover {
```
iampava, line 12: It feels a bit strange to rotate the entire icon * rectangle.

1. Let's rotate just the white rectangle (without the icon) and just to `90deg` so it's straight.
2. Let's shorten the transition time. Make it smaller so it feels more responsive

client/components/Landing/LandingStats/CounterCard/CounterCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+import styles from './CounterCard.module.scss';
+
+interface Props {
+  icon: IconProp;
+  counter: number;
+  description: string;
+  withAnimation?: boolean;
+}
+
+export default function LandingStats({
```
iampava, line 15: This should be named the same as the file - `CounterCard`

client/components/Landing/LandingStats/CounterCard/CounterCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+      {({ countUpRef }) => (
+        <div className={`${styles['counter-card']} text-center`}>
+          <div
+            className={`${styles['icon-box']} d-flex justify-content-center align-items-center`}
+          >
+            <FontAwesomeIcon width="30" height="30" color="#fff" icon={icon} />
+          </div>
+          <div className={styles['counter-details']}>
+            <p ref={withAnimation && countUpRef}>{counter}</p>
+            <h3>{description}</h3>
```
iampava, line 32: Judging from the UI, I'd say they should be inversed (the number is bigger and first).

However, I think here we should not use headings. We don't actually have a section following (like we would in an article) so maybe 2 paragraphs?

client/components/Landing/LandingStats/LandingStats.tsx
```diff
@@ -0,0 +1,52 @@
...
+  participants, events, trainers, hoursOfTraining, slides, countFrontEndWord,
+} : Props) {
+  return (
+    <div className={`${styles['landing-stats']}`}>
+      <section>
+        <h2 className="text-center">
+          Statistici
+        </h2>
+        <ul className="d-flex justify-content-center align-items-center">
+          <li>
```
iampava, line 30: We should have an array passed as a prop with all this info, and here we should do a `.map` over it.

This way:
1. We get rid of manually having to insert things here (ie: duplicate code)
2. We can re-use this component in other pages, but with different stats

pages/index.tsx
```diff
@@ -20,14 +20,20 @@ export default function Home() {
         description="Vrei să înveți FrontEnd? Aici ai parte de tutoriale gratuite și o comunitate de developeri care te vor ajuta să devii mai bun."
       />
       <>
-        {/* <LandingAdBanner /> */}
         <LandingHero />
+        <LandingStats
```
iampava, line 24: Let's move at the end of the home page.

Let's also add it to the events pages (at the end again)

iampava (changes requested):
Let's also update the numbers to the real ones and the texts a little bit 
* 11 FRONTEND-ERI CURIOȘI
* 2 WORKSHOP-URI PRACTICE
* 3 TRAINERI PASIONATI/EXPERIEMENTATI? (ninja is overused in a lot of websites/conferences so I'd keep away from it 😅)
* 7 ORE DE TRAINING
* 38 SLIDE-URI

Also, the numbers are being counted right at the beginning. So if you wait a little before scrolling you won't see that behavior.

Let's use IntersectionObserver and only start the counting when the user scrolls to that component.

iampava:
After you fix those small issues, I have another small idea to make  this even nicer.
But first let's implement what we know we're going to keep, and try this idea on a separate commit at the end (thus, if we don't need it we just remove that commit)

Since we're animating the numbers, what if we're animating the white rectangles as well? If we change them to SVG we can use this technique - https://codepen.io/janverstr/pen/PvPoaV?editors=1010 - and have the "white border" kinda like draw itself at the same time as the numbers are being counted.

Have a look at the codepen example, and this property: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray. You should try to adapt it for rects (not circles as in the example).

iampava (approved):

PR #136, opened 2021-03-30
Define components in StoryBook

andreeatoma:


client/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const Grey = () => (
+  <Button variant="grey"> Button </Button>
+);
+
+export const Bouncy = () => (
+  <Button variant="bouncy"> Button </Button>
+);
+
+export const WithIcon = () => (
+  <Button variant="with-icon"> Button </Button>
```
iampava, line 39: Typo here. The variant is called `with--icon`. 

Also, you should add an icon to this, to have a real example. If you wanna see where this button is used and how it's supposed to look like, you can look into `SocialMediaButtons.tsx`.

PS: you may need to add `withIcon` as a prop instead of as a `variant` since we can have **blue**, **success**, **danger** buttons with variants.

iampava, line 39: Typo here. The variant is called `with--icon`. 

Also, you should add an icon to this, to have a real example. If you wanna see where this button is used and how it's supposed to look like, you can look into `SocialMediaButtons.tsx`.

PS: you may need to add `withIcon` as a prop instead of as a `variant` since we can have **blue**, **success**, **danger** buttons with variants.

client/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+);
+
+export const OutlineDanger = () => (
+  <Button variant="outline btn--danger"> Button </Button>
```
iampava, line 43: Typescript should flag this:

image

The variant is wrong. Also, we're gonna need to add `outline` as a prop rather than as a variant since we should be able to have any button variant but with outline.

client/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+);
+
+export const Loading = () => (
+  <Button variant="loading"> Button </Button>
```
iampava, line 47: No `loading` variant here as well. TS should have flagged this.

client/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+import Button from './Button';
+
+export default {
+  title: 'Button',
+};
+export const Default = () => (
+  <Button variant="transparent"> Button </Button>
+);
+export const Primary = () => (
+  <Button variant="blue">  Button </Button>
```
iampava, line 11: The idea behind storybook is to see all your components and how you can customize them.

Here for the Primery/Blue buttons we have only the normal styling. What about `loading` and `disabled`?
For each variant we want to add these as well, so you see the entire picture.

Something like:
```
export const Primary = () => (
  <>
    <Button variant="blue">  Blue Button</Button>
    <Button variant="blue" disabled>  Disabled Blue Button</Button>
    <Button variant="blue" loading>  Loading Blue Button</Button>
  </>
)
```

client/components/Landing/LandingStats/LandingStats.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+  StatsCardLayout5,
+  StatsCardLayout6,
+} from '~/components/PlatformStats/v2/StatsCards';
+
+export default {
+  title: 'StatsCards',
+};
+
+export const StatsCard1 = () => (
+  <StatsCardLayout1
```
iampava, line 16: Now I see what you meant regarding the styles. It seems the cards have certain styles applied from the parent (`LandingStats`).

This is bad design and I should have caught it when we reviewed them. Not a big deal then, we're gonna fix here.

So, we remove all the card-related styles  from `LandingStats.module.scss`. It should only contain positioning styles for the cards. Thus these should be moved to Background 👇
1. the `padding`
2. the `before`

This way the cards in Storybook will look like the ones in our main app, apart from the positioning.

client/components/Landing/LandingStats/LandingStats.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const StatsCard5 = () => (
+  <StatsCardLayout5
+    text1="Fondator"
+    text2="de la comunitatea open-source"
+    colorText="#fff"
+  />
+);
+
+export const StatsCard6 = () => (
+  <StatsCardLayout6
```
iampava, line 63: Some cards have configurable properties, like this one. Just like for the buttons, we should have multiple cards here to show all possible variants. If we don't show everything that's possible, we might go into Storybook and think something doesn't exist already and we have to build it. But it does. 

So, for example in this card. The `footer` is optional. We should have a version without it.

Also, `background` can be an image. Let's have a version with that as well. 

PS: double check this for all cards, others may have optional props.

client/components/Landing/LandingStats/LandingStats.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    counter2={33}
+    text1="slide-uri"
+    colorText="#1b1b15"
+    text2="challenge-uri practice"
+    title="Am creat și"
+  />
+);
+
+export const StatsCard5 = () => (
+  <StatsCardLayout5
```
iampava, line 55: This card is specific for contributors, as it accepts the `contributors` prop (none other do).

However you didn't add this prop 😅

iampava (changes requested):

PR #144, opened 2021-04-15
[Exercises] added content on page

andreeatoma:


iampava:
I think this is a little too much content, plus it has a lot of UI elements and makes it seem like they were randomly chosen:

1. we have the main heading first
2. then the `# text` 
3. then the blue-bar text
4. and finally 4 more paragraphs

Way way too much things going on 😅

image

Let's try and resume everything in just a few sentences. The main points we need to touch are:

1. That this is the exercises page (I guess it's clear from the Heading)
2. That we believe getting feedback is really helpful
3. That we will give you feedback on those exercises if you send them
4. That we do this as a community project, for free
5. Link to the "About us" page which has more info

andreeatoma:
image

PR #147, opened 2021-04-20
[BUGFIXES] resolved errors at build

andreeatoma:


client/components/DevToolsClone/DevToolsClone.tsx
```diff
@@ -116,8 +116,12 @@ const DevToolsClone = () => {
       </section>
       <footer className={styles.footer}>
         <p>
-          Lățimea și înălțimea finală:{' '}
-          <strong>{getFinalSize()}px</strong>
+          Lățimea și înălțimea finală:
```
iampava, line 119: I wanted to merge this but there are soooo  many modifications, most of which are related to indentation / code formatting. It's hard to follow and find what are the changes that impact the build, and what are the esthetic changes.

I think we might have slightly different configs for ESLint or something like that 🤔 Anyway, that doesn't really matter for this PR, it should have "just" the modifications that FIX the build. No change in indentation, spacing, etc.

In other words, revert everything, precisely modify just what needs to be modified to fix the build, then save and make sure only that changed. If your editor is formatting on save (I suspect it does) - then you can choose the `Save without formatting` option from the command palette (Ctrl + Shift + P).

iampava, line 119: I wanted to merge this but there are soooo  many modifications, most of which are related to indentation / code formatting. It's hard to follow and find what are the changes that impact the build, and what are the esthetic changes.

I think we might have slightly different configs for ESLint or something like that 🤔 Anyway, that doesn't really matter for this PR, it should have "just" the modifications that FIX the build. No change in indentation, spacing, etc.

In other words, revert everything, precisely modify just what needs to be modified to fix the build, then save and make sure only that changed. If your editor is formatting on save (I suspect it does) - then you can choose the `Save without formatting` option from the command palette (Ctrl + Shift + P).

iampava (changes requested):

iampava:
@andreeatoma is this still needed? Is your build still crashing?

iampava:
Closing. Stale

PR #149, opened 2021-04-21
[Lesson] Links and buttons

iampava:


client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+            <i>ancoră</i>
+            {' '}
+            sau
+            {' '}
+            <i>link</i>
+            {' '}
+            este unul din cele mai uzuale elemente
+            folosite pentru construirea paginilor web. Dacă vrem să ne mutam la
+            o altă pagină sau la o altă zonă în cadrul aceleiași pagini, atunci
+            vom folosi acest element. Modalitatea prin care adăugăm un link
```
iampava, line 96: We don't add the link via the `href` attribute, we specify where the link points (destination)

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+          <LessonHeading as="h4" id="cele-mai-uzuale">
+            Cele mai uzuale
+          </LessonHeading>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a href="https://frontend.ro/"> Frontend.ro </a>`}
+          />
+          <LessonFigure
```
iampava, line 118: Move to an "inline" demo, meaning you just add the link in the article text and specify (maybe with bold text) "This is an absolute link". Remove the demo page for absolute links, I think it's pretty useless.

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+            {' '}
+            <strong>relative</strong>
+            , de obicei folosite
+            la realizarea meniurilor paginilor web:
+          </p>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a href="/team.html">Team</a>`}
```
iampava, line 144: Relative links are not well explained. This > `/team.html` is not a relative link, but an absolute one. It will always start from the "root" and find the "team.html" page there.

To be a relative link it must lead us to a different destination, depending on our current location. 

This is a relative link `team.html` (without `/`). Also this `./team.html` or this -> `../team.html`.

For the demo, I think we should show a simple folder structure with 2 foldrse: `home` and `settings`. In `home` let's have a `home.html` and `logo.jpg`. Also in `settings` let's have a `settings.html` and a `logo.jpg`. If we add this link `<a href="logo.jpg"> open image in another tab </a>` to both of them, it will open up a different image.


PS: please read the full article again, and change all explanations related to relative links

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+            withBorder
+            src="/images/lessons/links-and-buttons/jump-links.png"
+            alt="Link-uri care duc către o secțiune ale
+            aceleiași pagini sau așa numitele jump links"
+            demo="/demo/html/jump-links"
+          />
+          <LessonTip>
+            Observați atributul
+            {' '}
+            <strong className="formatted">ID</strong>
```
iampava, line 179: `id`  -> lowercase

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+            className="my-5"
+            language="html"
+            code={`
+<a href="#Section-further-down"> 
+  Sari la secțiunea de mai jos 
+</a> 
+<h2 id="Section-further-down">Section further down</h2>
+            `}
+          />
+          <LessonFigure
```
iampava, line 169: Pretty useless demo when compared to the video below. Let's remove and keep the video. 

When clicking the video it should navigate you to the demo page (pattern already exists in other articles) which should actually have stuff you jump to. Right now the demo page doesn't do anything, you click the link and nothing happens because there's no scroll. People might think it's broken.

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+          </p>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a href="mailto:hello@frontend.ro">
+  Trimite email la Frontend.ro
+</a>`}
+          />
+          <LessonFigure
```
iampava, line 221: Inline demo (like for the absolute URLs). Also show a video from mobile of how this works ( I think there are apps to record mobile screen. On iphone it can be done directly from the OS)

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+          </p>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a href="tel:+49.157.0156">+49 157 0156</a>
+<a href="tel:+1(555)5309">(555) 5309</a>
+`}
+          />
+          <LessonFigure
```
iampava, line 239: Inline demo and show video of example from mobile.

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+            deschidă în pagina curentă:
+          </p>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a href="/files/test.pdf" download>Download PDF</a>
+`}
+          />
+          <LessonFigure
```
iampava, line 260: Inline demo. (pretty much inline all demos and possibly show videos for a couple that are more interesting. Having all those pictures with just text here takes up a lot of space and don't really do anything)

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a href="http://frontend.ro" target="_blank" rel="noopener noreferrer">
+  Link-ul se deschide într-un nou tab
+</a>
+            `}
+          />
+          <LessonFigure
```
iampava, line 324: Remove demo completely, similar with previous ones. Plus the difference is actually from a security point.

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+          </p>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a title="Nu ar trebui să fiu aici" href="/">
+  Un simplu exemplu
+</a>`}
+          />
+          <LessonFigure
```
iampava, line 377: Replace with a more relevant picture in which we actually show how "title" looks when hovered. Get rid of the demo page, and inline this link.

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+  aria-hidden="true" focusable="false">
+  <circle cx="16" cy="16" r="16"/>
+  </svg>
+  <!--Se comporta ca un label ce e ascuns -->
+  <span style="position: absolute; opacity: 0">
+  Useful link text
+  </span>
+ </a>`}
+          />
+          <LessonFigure
```
iampava, line 420: Inline demo. I think we can keep the demo page, this is actually a little more interesting.

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+          </p>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<a href="/">
+  <img src="images/logo.png" alt="Frontend.ro LOGO." />
+</a>`}
+          />
+          <LessonFigure
```
iampava, line 459: Inline demo. We can keep the demo page if we want, but let's make it more "complex" Maybe combine links in pictures with some other features.???

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+        <section>
+          <LessonHeading as="h3" id="button-element">
+            {'<button> element'}
+          </LessonHeading>
+          <p>
+            Elementul html
+            {' '}
+            <strong className="formatted">{'<button>'}</strong>
+            {' '}
+            este folosit atunci când vrem să facem o acțiune (exemple uzuale de
```
iampava, line 497: Remove "exemple uzuale de actiuni" and keep just "exemple". You just mentioned actions.

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+            acțiuni: selectăm un element sau mergem cu mouse-ul deasupra lui
```
iampava, line 498: What does "selectam un element" mean? what is this action, I am a bit confused?

client/curriculum/html/LinksAndButtons.tsx
```diff
@@ -0,0 +1,882 @@
...
+            (hover) , o pagină web se încarcă (load), dăm click pe un buton sau
+            alegem o cheie de la tastatură, facem resize la o fereastră sau o
+            închidem, dăm play sau pause la un video, etc.
+          </p>
+          <p>
+            Spre deosebire de link-uri, butoanele
+            {' '}
+            <strong>NU</strong>
```
iampava, line 506: I think we should expand on this some more. We DON'T WANT to change the page directly, that is a links job. However, if the page changes due to a button action (click login button > wait > change page) then it's fine

PR #214, opened 2021-06-18
Adaugare FlexboxDefense

FrancuRazvan:
Sper ca e ok ce am facut


client/components/learn/Resources/resources-model.ts
```diff
@@ -150,6 +150,11 @@ export const RESOURCES: Record<ResourceCategory, ResourceDescription[]> = {
       url: 'https://flexboxfroggy.com/',
       cover: `${process.env.CLOUDFRONT_PUBLIC}/public/images/resources/flexboxfroggy_500w.png`,
     },
+    {
+      title: 'Flexbox Defense',
+      url: 'http://www.flexboxdefense.com/',
+      cover: `https://www.flickr.com/photos/193258951@N02/`,
```
iampava, line 156: Almost perfect! Just a few things regarding the image need changing:

1. Let's use the official image: http://www.flexboxdefense.com/images/gameboard-b58cd885d125374f5156eac2d6c3b62c.png

2. Let's download the image in our project, and save it in the `public/images/resources` folder. This way we make sure it is always available, even if FlexboxDefense somehow removes the image.

3. As you can see there, all images follow a naming convention, having the width in their name. For this one, let's resize it to 500px (you can use Paint if you're on Windows) and then rename it to `flexbox-defense_500w`

4. Use the source of the local image here. In other words, the string should be something like `/images/resources/flexbox-defense_500w.png/jpg`

After this we can merge it :D

iampava (changes requested):
I'm gonna give the feedback in English, in case other people from GitHub stumble upon our repository, they can know what's happening.

client/components/learn/Resources/resources-model.ts
FrancuRazvan: oky, done it...i hope i did it well

iampava: Almost 😅 The `cover` property is perfect! However, the image is not in the commit. 

You should download, rename then commit the image tother with the rest of the code.

FrancuRazvan: Done.....sorry i forgot to commit the img itself...:)

iampava: No worries :D 

One last question, do you know how to squash commit - meaning to combine several commits into just one? 

The code is great but we have 3 commits for one fix. We should have just one commit, named in English, something like:

`[Resources] Added Flexbox Defence`

PR #217, opened 2021-06-20
[Lesson] About HTML

catalinpopusoi:


client/curriculum/html/AboutHtml.tsx
```diff
@@ -0,0 +1,335 @@
...
+            <strong className="formatted">{'<section>'}</strong>
+            {' '}
+            are drept continut alte doua elemente (
+            <strong className="formatted">{'<h1>'}</strong>
+            {' si '}
+            <strong className="formatted">{'<p>'}</strong>
+            ), in timp ce aceste doua elemente au in interior continut
+            de tip text.
+          </p>
+          <LessonTip>
```
iampava, line 152: Self-closing elements feel pretty important to me so I wouldn't just use a `<LessonTip>` for them.

I would add a separate short paragraph and the code.

client/curriculum/html/AboutHtml.tsx
```diff
@@ -0,0 +1,335 @@
...
+          </LessonHeading>
+          <p>
+            Orice fisier HTML este construit din elemente, numite si
+            <strong>tags</strong>
+            . Ele sunt formate din perechi de forma
+            {' '}
+            <strong className="formatted">{'<element> </element>'}</strong>
+            . Primul se numeste
+            {' '}
+            <strong>opening tag</strong>
```
iampava, line 122: Let's use the romanian naming, and in paranthesis we can add the english one.

"tag de deschidre (opening tag)".

client/curriculum/html/AboutHtml.tsx
```diff
@@ -0,0 +1,335 @@
...
+            Ce este HTML?
+          </LessonHeading>
+          <p>
+            <strong>HTML (Hyper Text Markup Language)</strong>
+            {' '}
+            este limbajul standard de markup folosit pentru a defini
+            structura de baza a unei pagini web. Nu este un limbaj de
+            programare propriu-zis, ci doar un limbaj descriptiv, bazat pe
+            {' '}
+            <em>XML</em>
```
iampava, line 106: Here maybe it's worth adding a `<LessonTip>`? Something like:

"Knowing XML si not required to learn HTML, but if you used it before you may be at an advantage"

client/curriculum/html/AboutHtml.tsx
```diff
@@ -0,0 +1,335 @@
...
+          <p>
+            In HTML, un element bec care este de
+            {' '}
+            <strong>tip LED</strong>
+            , de
+            {' '}
+            <strong>culoare albastra</strong>
+            , ar putea fi scris dupa cum urmeaza:
+          </p>
+          <Highlight
```
iampava, line 192: Great analogy! 🥳

client/curriculum/html/AboutHtml.tsx
```diff
@@ -0,0 +1,335 @@
...
+<h1>Acesta este un titlu</h1>
+<p>Acesta este un paragraf</p>
+`}
+          />
+          <p>Acum, vom adauga si cateva atribute pe aceste elemente:</p>
+          <Highlight
+            className="my-5"
+            language="html"
+            code={`
+<h1 id=“main-title”>Acesta este un titlu</h1>
```
iampava, line 225: I think we should change what attributes we demo, and use things that actually make a difference directly.

`id` and `class` by themselves don't do anything, so I worry it might be hard for people that may run the demos on their machine to understand what is going on.

iampava (commented):
Overall looks great! Added some minor comments.

I also think we should at somewhere (maybe at the end) a suggestion like:

"Most people think HTML is very easy and take it lightly. That is partially true. HTML is easier than the other two languages, but it should be learned thoroughly. If you do so from the start, you'll have a great foundation from which to build you  CSS and JS skills"

client/curriculum/html/AboutHtml.tsx
```diff
@@ -0,0 +1,335 @@
...
+export default function AboutHtmlLesson() {
+  return (
+    <>
+      <SEOTags
+        title="Despre HTML | Lecție HTML"
+        shareImage="TBD"
+        description="Învață principiile de bază ale acestui limbaj și scrie primele linii de cod."
+        url="https://FrontEnd.ro/html/despre-html"
+      />
+      <Lesson id="despre-html" title="Despre HTML" chapters={chapters}>
```
iampava, line 35: Add `withExercises={false}` attribute. We won't have them for this language.

iampava (approved):

PR #265, opened 2021-08-03
Create a generic Responsive Flex component

andreeatoma:
Create a generic Responsive Flex component

remove unused code

### DIFF
--- a/client/components/ResponsiveFlex/ResponsiveFlex.stories.tsx
+++ b/client/components/ResponsiveFlex/ResponsiveFlex.stories.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import ResponsiveFlex from './ResponsiveFlex';
+
+export default {
+  title: 'ResponsiveFlex',
+};
+
+export const Default = () => (
+  <>
+    <ResponsiveFlex breakpoint={1024}>
+      <div>
+        <img
+          src="https://FrontEnd.ro/logo.png"
+          alt="FrontEnd.ro logo"
+        />
+      </div>
+      <div>
+        <h2>
+          Ce e FrontEnd.ro?
+        </h2>
+        <p>
+          FrontEnd.ro este o comunitate de oameni ce învață FrontEnd de la zero.
+        </p>
+        <p>
+          Comunitatea gravitează în jurul unei serii de lecții și exerciții gratuite
+          ce constituie - în opinia noastră - fundamentele unei cariere moderne
+          în FrontEnd Development.
+        </p>
+      </div>
+    </ResponsiveFlex>
+  </>
+);
--- a/client/components/ResponsiveFlex/ResponsiveFlex.tsx
+++ b/client/components/ResponsiveFlex/ResponsiveFlex.tsx
@@ -1,0 +1,59 @@
+import React, { PropsWithChildren, useEffect } from 'react';
+
+interface Props {
+  breakpoint: number;
+  className?: string;
+}
+
+// This component allows us to have dynamic, responsive layouts.
+// The component receives the breakpoint as props and
+// then switch from flex-row to flex-column.
+function ResponsiveFlex({ breakpoint, children, className = '' } : PropsWithChildren<Props>) {
+  const containerRef = React.createRef<HTMLDivElement>();
+  const responsiveClassName = `responsive-flex--${breakpoint}`;
+
+  // For each breakpoint value we create a className with the rules specific to that breakpoint.
+  // we first check if we have a unique class on the container and if we don't,
+  // we write the @media CSS code and create a generic class such as
+  // (responsive-flex--1000 or responsive-flex--256) that we assign to container
+  // otherwise we simply assign the generated class to our container
+  useEffect(() => {
+    if (!classExists(breakpoint)) {
+      const styleEl = createStyleEl(breakpoint);
+      document.head.appendChild(styleEl);
+    }
+  }, [breakpoint]);
+
+  function classExists(breakpoint): boolean {
+    return !!document.head.querySelector(`[data-flex-breakpoint="${breakpoint}"]`);
+  }
+
+  function createStyleEl(breakpoint) {
+    const styleEl = document.createElement('style');
+    styleEl.setAttribute('data-flex-breakpoint', `${breakpoint}`);
+    styleEl.innerHTML = `
+    .responsive-flex--${breakpoint} {
+      display: flex;
+      flex-direction: row;
+    }
+    @media screen and (max-width: ${breakpoint}px) {
+      .responsive-flex--${breakpoint} {
+        flex-direction: column;
+      }
+      .responsive-flex--${breakpoint} > * {
+        width: 100%;
+        margin-left: 0;
+        margin-right: 0;
+      }
+    }`;
+    return styleEl;
+  }
+
+  return (
+    <div ref={containerRef} className={`${responsiveClassName} ${className}`}>
+      {children}
+    </div>
+  );
+}
+
+export default ResponsiveFlex;
--- a/client/curriculum/html/HTMLStructure/HTMLStructure.tsx
+++ b/client/curriculum/html/HTMLStructure/HTMLStructure.tsx
@@ -1,4 +1,4 @@
-import React, { HTMLAttributes, PropsWithChildren } from 'react';
+import React from 'react';
 import SEOTags from '~/components/SEOTags';
 import Lesson, {
   LessonContributors,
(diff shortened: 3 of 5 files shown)


client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,45 @@
...
+  breakpoint: number;
+}
+
+function ResponsiveFlex({ breakpoint, children }: PropsWithChildren<Props>) {
+  const containerRef = React.createRef<HTMLDivElement>();
+
+  useEffect(() => {
+    if (breakpoint) {
+      if (document.head.querySelector(`[data-flex-breakpoint="${breakpoint}"]`)) {
+        containerRef.current.className = `responsive-flex--${breakpoint}`;
```
iampava, line 13: We shouldn't set the class like this (via traditional DOM methods). Instead at the begging of the function we can directly have `const responsiveClassName = 'responsive-....'`  (because we know what class this component will have depending on the breakpoint).

And when rendering `<div ref={containerRef} className={responsiveClassName }>`.

PS: let's also allow this component to receive classes as props. For example, we want to be able to do this

```
<ResponsiveFlex breakpoint={500} className="mt-2 mb-4">
```

and the rendered container should have both the breakpoint-related class and `mt-2 mb-4`.

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,45 @@
...
+            width: 100%;
+            margin-left: 0;
+            margin-right: 0;
+          }
+        }`;
+        document.head.appendChild(styleEl);
+        containerRef.current.className = `responsive-flex--${breakpoint}`;
+      }
+    }
+  }, [containerRef]);
```
iampava, line 36: Why do we have the `containerRef` as a dependency for `useEffect`?

I think we need to have the `breakpoint`, no?

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,45 @@
...
+interface Props {
+  breakpoint: number;
+}
+
+function ResponsiveFlex({ breakpoint, children }: PropsWithChildren<Props>) {
+  const containerRef = React.createRef<HTMLDivElement>();
+
+  useEffect(() => {
+    if (breakpoint) {
+      if (document.head.querySelector(`[data-flex-breakpoint="${breakpoint}"]`)) {
```
iampava, line 12: Let's try and make this a little more readable by moving this code into a separate function `classExists(breakpoint): boolean` that we'll call.

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { PropsWithChildren, useEffect } from 'react';
+
+interface Props {
+  breakpoint: number;
+}
+
+function ResponsiveFlex({ breakpoint, children }: PropsWithChildren<Props>) {
```
iampava, line 7: Please document this component. It's quite complex so let's write a little bit about why we want it and how it works.

Also, let's add it to Storybook.

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const containerRef = React.createRef<HTMLDivElement>();
+
+  useEffect(() => {
+    if (breakpoint) {
```
iampava, line 11: Why do we need this `if`?

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,45 @@
...
+        @media screen and (max-width: ${breakpoint}px) {
+          .responsive-flex--${breakpoint} {
+            flex-direction: column;
+          }
+          .responsive-flex--${breakpoint} div {
+            width: 100%;
+            margin-left: 0;
+            margin-right: 0;
+          }
+        }`;
```
iampava, line 31: Let's create a separate function `createStyleEl(breakpoint)` that will have all this above. The code should read:

```
if (!classExists(breakpoint)) {
  const styleEl = createStyleEl(breakpoint);
  document.head.appendChild(styleEl);
}
```

This will be easier to read by other devs, instead of having all that code inlined.

client/curriculum/html/HTMLStructure/HTMLStructure.tsx
```diff
@@ -369,17 +370,4 @@ function HTMLStructure() {
   );
 }
 
-type ResponsiveFlexProps = PropsWithChildren<HTMLAttributes<HTMLDivElement>>;
```
iampava, line 372: Good job deleting this.

Let's also delete the related styles from `HTMLStructure.module.scss`.

iampava (changes requested):
Overall great, the code works, good job 👍

Left some suggestions regarding making the code more readable.

client/curriculum/html/HTMLStructure/HTMLStructure.tsx
andreeatoma: Here I left the 
```
.main-structure-wrapper {
  width: 20em;
  margin-left: 5em;
}
```
because I can see that it's used here in the component:
```
 <Highlight
              language="html"
              className={styles['main-structure-wrapper']}
 />
```

client/components/ResponsiveFlex/ResponsiveFlex.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+import Highlight from '../Highlight/Highlight';
+import ResponsiveFlex from './ResponsiveFlex';
+
+export default {
+  title: 'ResponsiveFlex',
+};
+
+export const Default = () => (
+  <>
+    <ResponsiveFlex breakpoint={1024}>
```
iampava, line 12: Please simplify the example. There are a lot of elements here and it's not clear what should I be looking at. I propose we do just a very simple example which is enough to illustrate what the component is doing:

```
<ResponsiveFlex ...>
  <div> ... </div>
   <div> ... <div>
```

Inside the Divs you can add some text or images if you want, but it shouldn't be very complicated.

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { PropsWithChildren, useEffect } from 'react';
+
+interface Props {
+  breakpoint: number;
+  className?: string;
+}
+
+// This is a generic component that receives the breakpoint and a className as props and
```
iampava, line 8: The initial sentence should be a generic "why?". Something like this: `This component allows us to have dynamic, responsive layouts`.

Then you can exaplain that it will receive a breakpoint and switch from `flex-row` to `flex-column`.
You can delete the `className` mention, that's not relevant enough to add in the comment.

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,61 @@
...
+// then switch from row to column layout.
+// Because this component will be used in multiple places, we wanted a simple implementation,
```
iampava, line 10: Delete this sentence. You're explaining below what;'s going on.

client/components/ResponsiveFlex/ResponsiveFlex.tsx
```diff
@@ -0,0 +1,61 @@
...
+// with very little js.
+
+function ResponsiveFlex({ breakpoint, children, className = '' } : PropsWithChildren<Props>) {
+  const containerRef = React.createRef<HTMLDivElement>();
+  const responsiveClassName = `responsive-flex--${breakpoint}`;
+
+  // we first check if we have a unique class on the container and if we don't,
```
iampava, line 17: Add this to the begining:

`For each breakpoint value we create a className with the rules specific to that breakpoint`.
 
I think it provides more clarity. Then you can continue with your explanation 👍  (we first check if the class already exists. If not ...)

client/curriculum/html/HTMLStructure/HTMLStructure.tsx
iampava: Great!

iampava (changes requested):
Almost there. Some small doc changes but the code is good 👍

Also, please rename the commit to have the issue number in it, so it's highlighted like the rest of the commits on master. 
image

iampava (approved):

Issue #259, opened 2021-08-02
Lists SHOULD NOT override the default styles


iampava:
Right now in `_lists.scss` we're overriding the default styles for lists.

THIS IS A BIG PROBLEM! 😅 When creating exercises via Markdown (for example) no lists are rendered as the styles get overridden.

We should have a generic list component with a `variant` prop through which we control styling. Example:

```
<List variant="none"> ... </List>
<List variant="checkmark"> ... </List>
...
```

PR #290, opened 2021-08-12
[BUGFIX] Lists SHOULD NOT override the default styles #259

andreeatoma:
replace ul element with generic List to see how it appears

### DIFF
--- a/client/components/List/List.tsx
+++ b/client/components/List/List.tsx
@@ -1,0 +1,45 @@
+import React, { PropsWithChildren } from 'react';
+import styles from './List.module.scss';
+
+interface Props {
+  variant?: 'none' | 'bullets' | 'checkmark' | 'count';
+  as?: 'ul' | 'ol';
+  className?: string;
+}
+
+const List = React.forwardRef<
+  HTMLUListElement | HTMLOListElement,
+  PropsWithChildren<Props> &
+    React.OlHTMLAttributes<HTMLOListElement | HTMLUListElement>
+>(
+  (
+    {
+      variant = 'none',
+      as = 'ul',
+      className = '',
+      children,
+      ...props
+    }: PropsWithChildren<Props> &
+      React.HTMLAttributes<HTMLUListElement | HTMLOListElement>,
+    forwardRef,
+  ) => {
+    if (as === 'ul') {
+      return (
+        <ul ref={forwardRef} className={`${styles[variant]} ${className}`} {...props}>
+          {children}
+        </ul>
+      );
+    }
+    return (
+      <ol
+        ref={forwardRef as React.ForwardedRef<HTMLOListElement>}
+        className={`${styles[variant]} ${className}`}
+        {...props}
+      >
+        {children}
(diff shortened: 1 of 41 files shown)


client/components/List/List.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { PropsWithChildren } from 'react';
+
+interface Props {
+  variant?:
+  | 'none'
+  | 'with--bullets'
```
iampava, line 6: The classes were written in this way to simulate some kind of BEM syntax - http://getbem.com/introduction/

However, being a prop we can get rid of `with--` and just leave the second part `none`, `bullets`, `checkmark`...

client/components/List/List.tsx
```diff
@@ -0,0 +1,30 @@
...
+  | 'with--checkmark'
+  | 'with--count',
+  type: 'ul' | 'ol',
```
iampava, line 9: Let's rename this to `as` so we are consistent with other Components (eg: `FormattedText`)

client/components/List/List.tsx
```diff
@@ -0,0 +1,30 @@
...
+  className?: string,
+}
+
+function List({
+  variant = 'none', children, type, className = '',
```
iampava, line 14: Good job also passing the className 💪

client/components/List/List.tsx
```diff
@@ -0,0 +1,30 @@
...
+} : PropsWithChildren<Props>) {
+  return (
+    type === 'ul' ? (
```
iampava, line 17: We can simplify this code and have something similar to the `FormattedText` component. Look there for inspiration.

PS: I think instead of `Wrapper` here we should name it `List` so it's clearer what is going on.

client/styles/_lists.scss
```diff
@@ -5,7 +5,7 @@ ol {
   list-style: none;
 }
 
-ul.with--bullets {
+.with--bullets {
```
iampava, line 8: Since we're building a new component, we should move these styles in that folder and use them the way we're doing for the rest:

```
import styles from ....;
```

iampava (changes requested):
Overall looks good! 👏

We can improve it a little bit, see my comments.

client/styles/_lists.scss
```diff
@@ -3,47 +3,4 @@ ol {
   margin: 0;
```
iampava, line 3: Delete this as well (i mean the entire file). This was the initial problem, we don't want to override the styles anymore.

Basically there should be no default styles for `ul` or `ol`.

PS: have another look at the List storybook after you delete this file and make sure it looks ok.

iampava (changes requested):

iampava (commented):
As talked offline (on Slack) please double check all usecases of `ul` or `ol` to make sure they look as before.

Also, when squashing/rebasing, comment the other lines in the commit (right now all of them appear)

image

iampava (approved):

PR #303, opened 2021-09-02
center hide/show password icon

lazarnarcis:


client/components/Form/FormGroup/FormGroup.module.scss
```diff
@@ -14,4 +14,10 @@
     width: 100%;
     display: block;
   }
+  input[type="text"]:focus,
```
iampava, line 17: This PR should only be related to centering the icon inside the `<PasswordReveal>` component. Let's revert this code.

client/components/Form/PasswordReveal/PasswordReveal.tsx
```diff
@@ -34,6 +34,9 @@ function PasswordReveal({
           className="absolute text-grey"
           onClick={() => setVisible(!visible)}
           title={visible ? 'Hide password' : 'Show password'}
+          style={{
```
iampava, line 37: It's not a good practice to use inline styles or negative margins. 

Have a look in `PasswordReveal.module.scss` and add modify the `top` property to have it centered.

iampava (changes requested):
First of all thanks for the commit I appreciate contributions to this project! 🙏

Unfortunately there are a few problems that we must address, on top of the feedback I left in the code:

1. Please create an issue describing the problem. In this case the issue can be named: "Center reveal icon inside PasswordReveal component"

2. Link this PR to that issue so that when we merge this PR, the issue is automatically closed

3. There should not be a `package-lock.json` file since we're using `yarn`. Delete it.

4. I see there are also modifications inside `yarn.lock`. Those should not be there - you didn't add any new packages. Please revert back to the previous version of this file.

5. Finally, after you finished everything (including the comments on the code) please **squash** into a single commit that is named the same as the issue

PR #499, opened 2022-01-02
Realign the password reset button (https://github.com/FrontEnd-ro/fro…

lazarnarcis:
…ntend.ro/issues/497)

### DIFF
--- a/client/components/login/LoginButtons/LoginButtons.module.scss
+++ b/client/components/login/LoginButtons/LoginButtons.module.scss
@@ -14,11 +14,11 @@
 }
 
 .register-mode {
-  button:first-of-type {
+  .login-button {
     transform: translateY(5em);
   }
 
-  button:nth-of-type(2) {
+  .register-button {
     transform: translateY(-5em);
   }
 }
--- a/client/components/login/LoginButtons/LoginButtons.tsx
+++ b/client/components/login/LoginButtons/LoginButtons.tsx
@@ -19,7 +19,7 @@ export default function LoginButtons({
       <Button
         variant="light"
         type={mode === 'register' ? 'button' : 'submit'}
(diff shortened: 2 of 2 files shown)


client/components/login/LoginButtons/LoginButtons.module.scss
```diff
@@ -14,7 +14,7 @@
 }
 
 .register-mode {
-  button:first-of-type {
+  button:first-of-type:not(.forgot-password) {
```
iampava, line 17: This works, but I think a better fix is to add classes on the Login and Register buttons, and then here only transform those.

This way, other buttons are unaffected.

client/components/login/LoginButtons/LoginButtons.tsx
```diff
@@ -38,7 +38,7 @@ export default function LoginButtons({
         Înregistrează-te
       </Button>
       <div className="text-right mt-4 text-xs">
-        <Button variant="link" type="button" onClick={onResetPassword}>
+        <Button variant="link" type="button" onClick={onResetPassword} className={`${styles['forgot-password']}`}>
```
iampava, line 41: You don't have to do: 
```
className={`${styles['forgot-password']}`}
```

Since it's just one, you can simply do: `className={styles['forgot-password']}`.

iampava (changes requested):

client/components/login/LoginButtons/LoginButtons.module.scss
iampava: So, if in the future we keep adding buttons to this UI, we won't need to make any changes.

lazarnarcis (commented):
Problem solved!

iampava:
You still have 2 commits...

lazarnarcis:
Now it's just a commit!

iampava (approved):

PR #504, opened 2022-01-02
Close menu using the esc key

lazarnarcis:


client/components/layout/AsideMenu/AsideMenu.tsx
```diff
@@ -34,6 +34,17 @@ function AsideMenu({
     } else {
       document.body.style.overflow = 'initial';
     }
+
+    const handleEsc = (event) => {
```
iampava, line 38: Let's extract this into a separate hook, so we can easily re-use this code among components.

Have a look at `useOutsideClick` for inspiration.

iampava (changes requested):

lazarnarcis (commented):
Problem solved!

client/services/Hooks.ts
```diff
@@ -99,11 +99,27 @@ function withAuthModal(isLoggedIn: boolean, cb: (...any) => any) {
   };
 }
 
+function pressEsc(ref: MutableRefObject<HTMLElement>, handler: (e: MouseEvent) => void) {
```
iampava, line 102: Couple of problems here:

1. Why do we need the `ref` parameter? It's not used at all.
2. The type `e: MouseEvent` is wrong. It should be a keyboard event.

And finally, let's rename to `useKeyDown` and make it generic, so it can accept any key. This way we can reuse it in different scenarios.

client/services/Hooks.ts
```diff
@@ -99,11 +99,27 @@ function withAuthModal(isLoggedIn: boolean, cb: (...any) => any) {
...
+  useEffect(() => {
+    const handleEsc = (event) => {
+      if (event.keyCode === 27) {
```
iampava, line 105: `keyCode` is deprecated - https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Let's use `code` instead.

iampava (changes requested):
Better, but we can still improve it further :)

lazarnarcis (commented):
Now is ok?

client/services/Hooks.ts
```diff
@@ -99,11 +99,27 @@ function withAuthModal(isLoggedIn: boolean, cb: (...any) => any) {
   };
 }
 
+function pressEsc(handler: (e: KeyboardEvent) => void) {
```
iampava, line 102: In [this comment](https://github.com/FrontEnd-ro/frontend.ro/pull/504#discussion_r777892740) i suggested naming it `useKeyDown`. This is a general convention among React hooks. Let's stick to it :)

client/services/Hooks.ts
```diff
@@ -99,11 +99,27 @@ function withAuthModal(isLoggedIn: boolean, cb: (...any) => any) {
...
+  useEffect(() => {
+    const handleEsc = (event) => {
+      if (event.code === 27) {
```
iampava, line 105: Did you test your code after the changes? `event.code` is not a number, but a string. Pretty sure this code doesn't work.

image

client/services/Hooks.ts
iampava: I also suggested to refactor the function to be generic. What I mean by that, is I should be able to use it for other keys as well, not just Escape. Example:

```
useKeyDown('Escape', doSmthOnEscape);

...

useKeyCode('Space', doSmthOnSpace)
```

Please refactor.

iampava (changes requested):
You missed some of my previous comments.

lazarnarcis (commented):
I think it's ok now :)

client/services/Hooks.ts
```diff
@@ -99,11 +99,27 @@ function withAuthModal(isLoggedIn: boolean, cb: (...any) => any) {
   };
 }
 
+function useKeyDown(handler: (e: KeyboardEvent) => void, eventName: string) {
```
iampava, line 102: `eventName` is not a good choice for this variable name. It makes me think of DOM events (eg: `onclick`, `onscroll`, etc). 

Let's rename to `code` and also add a comment here explaining this refers to the `KeyboardEvent.code` property.

Also, please reorder the variables, have `code` first.

client/services/Hooks.ts
```diff
@@ -99,11 +99,27 @@ function withAuthModal(isLoggedIn: boolean, cb: (...any) => any) {
...
+      if (event.code === eventName) {
+        handler(event);
+      }
+    };
+    
+    window.addEventListener('keydown', handleEsc);
+    return () => {
+      window.removeEventListener('keydown', handleEsc);
+    };
+  }, [handler]);
```
iampava, line 114: I just realized we have this`handler` here. Since it's a function, it will make this effect run on every re-render. Here's an example: https://codesandbox.io/s/new-tree-lvfe8?file=/src/Demo.js

Let's change it to:

```
 }, [code]);
```

iampava (changes requested):
Almost almost there 😃

iampava (approved):

PR #939, opened 2023-04-02
Convert the Lists lesson to MDX (#936)

damdiana:


client/services/MDXService.ts
```diff
@@ -76,6 +76,11 @@ class MDXService {
         mdxContent = mdxAsString as unknown as string;
         break;
       }
+        case 'liste': {
```
iampava, line 79: This is not alligned

iampava (approved):

